PR #6, opened 2021-04-21
Merge upstream Ruby

maximecb:


maximecb:
@XrXr I saw this MJIT failure on another PR as well. Wondering if that's somehow due to us?

Is YJIT automatically disabled when MJIT is enabled?

XrXr:
> Is YJIT automatically disabled when MJIT is enabled?

No. These MJIT failures look like the ones from upstream though.

maximecb:
@XrXr you've been seeing this failure recently?

PR #9, opened 2021-04-22
Make gen_opt_aref() delegate to gen_opt_send_without_block()

XrXr:
Basically mirroring the interpreter for the general case.
Gets about 1% more coverage in railsbench.


maximecb:
@XrXr wow I'm kind of amazed at how simple this PR is. 5 lines of code *doge wow*

PR #10, opened 2021-04-23
Implement greedy versioning. Refactor versioning logic.

maximecb:
Testing out a hypothesis here. I realized that the way we allocate block versions in YJIT is different than Higgs. YJIT takes the first match that we can find, even if it's not perfect. However, Higgs tries to generate perfectly matching versions until the version limit is hit.

The downside with the "conservative" versioning approach taken by YJIT is that when it comes to passing argument/self types to callees, if the function has ever been compiled with a generic context (no type information), for example because it's been called by the interpreter, then no type information can ever be passed to this function through a specialized context. This is because the generic context will always match any incoming context, no matter what type information it carries.

So, in this PR, I implemented the more greedy versioning strategy taken by Higgs, where we try to generate perfectly matching versions that preserve type information until the limit is hit. Testing locally, the code size is actually smaller with greedy versioning (!?), on both railsbench and optcarrot.

I will try benchmarking for performance on AWS next. Let's also hope that the tests pass. PR also introduces a number of changes that I think make the versioning logic cleaner and easier to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,4 +96,15 @@ The core of CRuby's interpreter logic is found in:
 
 ## Contributing
 
-We welcome open source contributors. If you are interested in contributing to this project, please contact Maxime Chevalier [(@Love2Code) via twitter](https://twitter.com/Love2Code) or by email (maxime.chevalierboisvert@shopify.com). You can also feel free to open new issues to report bugs or just to ask questions. Suggestions on how to make this readme file more helpful for new contributors are most welcome.
+We welcome open source contributors. You should feel free to open new issues to report bugs or just to ask questions.
+Suggestions on how to make this readme file more helpful for new contributors are most welcome.
+
+Bug fixes and bug reports are very valuable to us. If you find bugs in YJIT, it's very possible be that nobody has reported this bug before,
+or that we don't have a good reproduction for it, so please open an issue and provide some information about your configuration and a description of how you
+encountered the problem. If you are able to produce a small reproduction to help us track down the bug, that is very much appreciated as well.
+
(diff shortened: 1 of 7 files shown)


maximecb:
So far no visible difference on AWS:
```
-------------  -----------  ----------  ---------  ----------  -----------  -------
bench          interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  1st itr
30k_ifelse     2359.0       0.0         420.8      0.1         5.61         4.33   
30k_methods    6487.4       0.0         1091.9     0.0         5.94         5.69   
activerecord   170.5        0.1         172.6      0.3         0.99         0.97   
binarytrees    428.5        1.9         380.1      2.4         1.13         1.13   
cfunc_itself   104.0        0.2         62.3       0.9         1.67         1.66   
fannkuchredux  4891.2       0.1         4906.7     0.1         1.00         1.00   
fib            237.2        0.0         76.2       0.2         3.11         3.11   
getivar        105.8        0.3         44.3       0.3         2.39         0.99   
lee            1264.4       0.9         1189.2     0.9         1.06         1.02   
liquid-render  201.0        1.0         181.3      1.1         1.11         1.10   
nbody          127.1        0.1         124.7      0.2         1.02         1.01   
optcarrot      6123.7       1.0         5161.0     0.2         1.19         1.18   
railsbench     3953.6       1.0         3961.2     1.0         1.00         0.98   
setivar        66.4         0.3         34.0       0.4         1.95         1.00   
-------------  -----------  ----------  ---------  ----------  -----------  -------
```

I will add some command-line options so we can experiment with enabling/disabling liberal versioning.

PR #11, opened 2021-04-25
[WIP] Take CFP from r15

jamatthews:
YARV with reg_cfp enabled already pins the CFP into R15 so I think we can continue that convention int YJIT and skip making the cfp a parameter for the fake call.

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -3396,7 +3396,7 @@ trace_set_i(void *vstart, void *vend, size_t stride, void *data)
 }
 
 void
-rb_yjit_empty_func_with_ec(rb_control_frame_t *cfp, rb_execution_context_t *ec)
+rb_yjit_empty_func_with_ec(rb_execution_context_t *ec, rb_control_frame_t *cfp)
 {
     // it's put in this file instead of say, compile.c to dodge long C compile time.
     // it just needs to be in a different unit from vm.o so the compiler can't see the definition
--- a/iseq.h
+++ b/iseq.h
@@ -301,7 +301,7 @@ VALUE rb_iseq_defined_string(enum defined_type type);
 /* vm.c */
 VALUE rb_iseq_local_variables(const rb_iseq_t *iseq);
 
-NOINLINE(void rb_yjit_empty_func_with_ec(rb_control_frame_t *cfp, rb_execution_context_t *ec));
+NOINLINE(void rb_yjit_empty_func_with_ec(rb_execution_context_t *ec, rb_control_frame_t *cfp));
 
 RUBY_SYMBOL_EXPORT_END
 
--- a/tool/ruby_vm/views/vm.inc.erb
+++ b/tool/ruby_vm/views/vm.inc.erb
@@ -32,7 +32,7 @@ INSN_ENTRY(yjit_call_example_with_ec)
 #if USE_MACHINE_REGS
     // assumes USE_MACHINE_REGS, aka reg_pc setup,
     // aka #define SET_PC(x) (reg_cfp->pc = reg_pc = (x))
-    rb_yjit_empty_func_with_ec(GET_CFP(), ec);
+    rb_yjit_empty_func_with_ec(ec, GET_CFP());
     RESTORE_REGS();
 #endif
     END_INSN(yjit_call_example_with_ec);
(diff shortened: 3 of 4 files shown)


XrXr:
It's not really a pin. It doesn't work on Clang:
This is the disassembly for `getclassvariable`
```
->  0x100289f23 <+4195>: mov    rax, r13
    0x100289f26 <+4198>: mov    r13, r8
    0x100289f29 <+4201>: mov    r12, rax
    0x100289f2c <+4204>: mov    rbx, qword ptr [rax + 0x8]
    0x100289f30 <+4208>: add    r12, 0x10
    0x100289f34 <+4212>: mov    qword ptr [r8], r12
    0x100289f37 <+4215>: mov    rax, qword ptr [r8 + 0x8]
    0x100289f3b <+4219>: mov    qword ptr [rax], 0x0
    0x100289f42 <+4226>: mov    edi, 0xa
    0x100289f47 <+4231>: call   0x1002df380               ; rb_yjit_collect_vm_usage_insn at yjit_iface.c:789
    0x100289f4c <+4236>: mov    rcx, qword ptr [r13 + 0x20]
    0x100289f50 <+4240>: mov    rdx, qword ptr [rcx]
    0x100289f53 <+4243>: test   dl, 0x1
    0x100289f56 <+4246>: je     0x100289fd8               ; <+4376> [inlined] VM_ENV_FLAGS + 5 at vm_core.h:1315
    0x100289f5c <+4252>: nop    dword ptr [rax]
    0x100289f60 <+4256>: mov    rax, qword ptr [rcx - 0x10]
    0x100289f64 <+4260>: test   dl, 0x2
    0x100289f67 <+4263>: jne    0x100289fdd               ; <+4381> [inlined] vm_env_cref + 141 at vm_insnhelper.c:828
    0x100289f69 <+4265>: test   rax, rax
    0x100289f6c <+4268>: je     0x100289fbf               ; <+4351> [inlined] VM_ENV_FLAGS at vm_core.h:1315
```
No mention of `r14` or `r15`.

And even on GCC it's explicitly unsupported to assume that it's always going to be in a certain register.

> The only supported use for this feature is to specify registers for input and output operands when calling Extended asm

jamatthews:
Sorry, this was just a half formed thought. @maximecb I was more thinking about reducing code size and load dependencies by using the callee-saved registers for state like the other fast interpreters and JITs with fixed register allocation like JavaScriptCore, Beam ASMJIT and LuaJIT do.

What do you think about shuffling the state into the callee-saved registers in yjit_entry_prologue? Should I give it a go and we'll see if it helps?

 [Erlang ASMJIT](http://erlang.org/documentation/doc-12.0-rc1/erts-12.0/doc/html/BeamAsm.html)
```
rbx: ErtsSchedulerRegisters struct (contains x/float registers and some metadata)
rbp: Active code index
r12: Optional Save slot for the Erlang stack pointer when executing C code
r13: Current running process
r14: Remaining reductions
r15: Erlang heap pointer
```

[LuaJIT](https://github.com/LuaJIT/LuaJIT/blob/75ee3a6159f1831fa57992df3caf5a2c303c281a/src/vm_x64.dasc#L30)
```
|.define KBASE,		r15		// Must be C callee-save.
|.define PC,		rbx		// Must be C callee-save.
|.define DISPATCH,	r14		// Must be C callee-save.
|.define KBASEd,	r15d
|.define PCd,		ebx
|.define DISPATCHd,	r14d
|.endif
```

[JavaScriptCore](https://github.com/WebKit/WebKit/blob/1df7d3ae8b3df2e05d40802f63bbf876f44167d5/Source/JavaScriptCore/offlineasm/x86.rb#L54)
```
# rbx =>             csr0 (callee-save, wasmInstance)
# r12 =>             csr1 (callee-save, metadataTable)
# r13 =>             csr2 (callee-save, PB)
# r14 =>             csr3 (callee-save, tagTypeNumber)
# r15 =>             csr4 (callee-save, tagMask)
# rsp => sp
# rbp => cfr
# r11 =>                  (scratch)
```

maximecb:
For now I think it isn't going to help because we're not doing enough C calls. We don't stay in the JIT long enough. However, as we shift to doing more calls, it could make a difference. That being said, we'll still need to save the CFP and SP registers often, in case C functions themselves make Ruby calls, throw exceptions, etc.

PR #16, opened 2021-04-28
Implement send with blocks

XrXr:
Not that much extra work compared to `opt_send_without_block`.
Moved the stack over flow check because it could've exited after changes
are made to cfp.


yjit_codegen.c
```diff
@@ -2032,6 +2047,11 @@ gen_opt_send_without_block(jitstate_t* jit, ctx_t* ctx)
         return YJIT_CANT_COMPILE;
     }
 
+    if (block && cme->def->type != VM_METHOD_TYPE_ISEQ) {
+        GEN_COUNTER_INC(cb, send_pass_block_to_weird_callee);
```
maximecb, line 2051: Maybe `send_block_to_not_iseq` ?

yjit_iface.c
```diff
@@ -481,6 +481,7 @@ rb_yjit_compile_iseq(const rb_iseq_t *iseq, rb_execution_context_t *ec)
 {
 #if OPT_DIRECT_THREADED_CODE || OPT_CALL_THREADED_CODE
     RB_VM_LOCK_ENTER();
+    // TODO: I think we need to stop all other ractors here
```
maximecb, line 484: You might be right. I guess if other ractors are running, they might start compiling the same code at the same time-ish?

yjit_iface.h
```diff
@@ -48,6 +48,8 @@ YJIT_DECLARE_COUNTERS(
     oswb_se_cc_klass_differ,
     oswb_se_protected_check_failed,
```
maximecb, line 49: We should probably rename all the `oswb_` counters to `send_`

maximecb:
Good job. It's great to finally support blocks! :) 

Is the test-bundled-gems failure here the same heisenbug you mentioned yesterday? :O

yjit_iface.c
XrXr: Yeah. Actually, I'm not sure if we need to stop all other ractors, but we should probably make sure only one ractor wins the race.

XrXr: Doesn't seem to be causing issue at the moment though. Maybe because we are not patching native code when generating entry points.

yjit_codegen.c
```diff
@@ -1738,22 +1757,30 @@ gen_oswb_cfunc(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const
...
     // Copy SP into RAX because REG_SP will get overwritten
     lea(cb, RAX, ctx_sp_opnd(ctx, 0));
 
+    // Pop the C function arguments from the stack (in the caller)
+    ctx_stack_pop(ctx, argc + 1);
+
+    // Write interpreter SP into CFP
+    lea(cb, REG_SP, ctx_sp_opnd(ctx, 0));
+    mov(cb, member_opnd(REG_CFP, rb_control_frame_t, sp), REG_SP);
+    ctx->sp_offset = 0;
```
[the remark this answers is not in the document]
XrXr: In case the C function calls the block, the block's frame inherits the sp. See commit message for b36c7f2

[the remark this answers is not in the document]
maximecb: But then should we have this under `if (block) {}` ?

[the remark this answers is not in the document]
XrXr: Hmm yeah. I like the idea of always having `ctx->sp_offset == 0` for the different ways of making calls, that way polymorphic call sites share the same successor. With it under `if(block)` we generate 55 bytes less code in `railsbench`. Don't know which way is faster but going with code size is probably a good first approximation.

PR #20, opened 2021-04-29
Implement dupn for n==2 (only case used in RDoc/railsbench)

maximecb:


XrXr (approved):
Love the data driven approach with finding out that it's only used for n==2.

PR #22, opened 2021-04-29
Add timeout for test-all tests when using parallel runner

XrXr:
Sometimes our CI timeout due to tests hanging. Set a timeout for tests
using the Timeout stdlib. The caveat is that if the test that hangs
doesn't check for Ruby interrupts the Timeout library might be
ineffective. Let's see put this on CI and see how it goes.

Pretty hacky code, I know, but hopefully it doesn't make CI less stable?
I tried using the Timeout stdlib at first, but some tests that look at
the backtrace fail if I wrap them in `Timeout.timeout`.

Alternative would be to just log the test names before running them, but
that would generate a lot of log data. Don't like that.

### DIFF
--- a/tool/lib/test/unit/parallel.rb
+++ b/tool/lib/test/unit/parallel.rb
@@ -49,6 +49,19 @@ module Test
           end
         end
 
+        # Thread for making sure the parallel worker is not stuck on one test
+        # for too long.
+        @timer_thread = Thread.new do
+          timeout = 8 * 60 # 8 minutes
+          while true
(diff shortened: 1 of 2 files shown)


tool/lib/test/unit/parallel.rb
```diff
@@ -49,6 +49,19 @@ def _run_suite(suite, type) # :nodoc:
           end
         end
 
+        @timer_thread = Thread.new do
```
maximecb, line 52: Looks good but you should comment your additions to explain what they do.

PR #47, opened 2021-05-21
Recommend installing libcapstone in README

jhawthorn:
@HParker and I were trying YJIT out and could have used this hint on how to make `YJIT.disasm` available


maximecb:
Thank you John! Let us know if you run into any issues or have any feedback you'd like to share :)

PR #48, opened 2021-05-23
Add newhash and newarray instructions to yjit codegen

jhawthorn:
I really enjoyed digging into yjit and couldn't resist adding some missing instructions 😁.

This adds `newhash` and `newarray`. `newarray` is fully implemented, but I only implemented `newhash` in the n=0 case (`{}`) though I don't think it would be too tough to implement the general case.

**`{}` (`newhash 0`) compiles to**:

```
  ; newhash
  55ffff57705d:  push   rdi
  55ffff57705e:  push   rsi
  55ffff57705f:  push   rdx
  55ffff577060:  push   rdx
  55ffff577061:  call   0x55fffd713260
  55ffff577066:  pop    rdx
  55ffff577067:  pop    rdx
  55ffff577068:  pop    rsi
  55ffff577069:  pop    rdi
  55ffff57706a:  mov    qword ptr [rdx], rax
```

**`[]` (`newarray 0`) compiles to**:

```
  ; newarray
  5642a2ac705d:  push   rdi
  5642a2ac705e:  push   rsi
  5642a2ac705f:  push   rdx
  5642a2ac7060:  push   rdx
  5642a2ac7061:  mov    rdi, rsi
  5642a2ac7064:  movabs rsi, 0
  5642a2ac706e:  lea    rdx, [rdx]
  5642a2ac7071:  call   0x5642a1a39d10
  5642a2ac7076:  pop    rdx
  5642a2ac7077:  pop    rdx
  5642a2ac7078:  pop    rsi
  5642a2ac7079:  pop    rdi
  5642a2ac707a:  mov    qword ptr [rdx], rax
```

The `lea rdx, [rdx]` is a bit of an odd coincidence 😅, but I figured at this point we shouldn't bother optimizing that away and favour readability/maintainability.

Let me know if there's any tests I should have or changes I can make. I'll take a look when I'm back in the office Wednesday ❤️


maximecb:
So the failing bundler test is a recurring issue we've been seeing (unsure the cause yet). The MJIT segfault I'm not sure about. Will try restarting the tests.

maximecb:
Still failing. Deeper investigation required into what causes the segfault. If you can't figure it out on your own we could try pairing tomorrow, or we could wait until Aaron is back from vacation next week.

maximecb:
Thank you :)

PR #49, opened 2021-05-23
Add putstring/opt_str_freeze/opt_str_uminus to yjit codegen

jhawthorn:
These instructions are more common in files not using `frozen_string_literal: true`, but Ruby generated from ERB still uses `.freeze` and slightly older/less heavily optimized code is likely to as well.

For YJIT, `opt_str_freeze` and `opt_str_uminus` are basically the same as `putobject`, we just need to check that their BOPs aren't redefined.

`putstring` needs to call resurrect to get a new string object.

Let me know anything I can change in this 😁


maximecb:
You might have better luck with the CI implementing just one instruction at once, or you can comment the other ones out. 
Otherwise it's hard to figure out what's not working 😮 

Would be good to have some new simple tests in `bootstraptest/test_yjit.rb`. Particularly if you can identify tricky corner cases.

But as I said on the other PR, if you can't figure out, we can help :)

maximecb:
Well done :)

PR #50, opened 2021-05-23
Detect types from putobject and getinlinecache

jhawthorn:
This implements the "TODO" in `gen_putobject` of performing more specific type detection. Doing so I realized we could also probably be doing the same getting constants in `opt_getinlinecache`.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -113,6 +113,34 @@ jit_peek_at_self(jitstate_t *jit, ctx_t *ctx)
     return jit->ec->cfp->self;
 }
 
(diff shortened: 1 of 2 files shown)


maximecb:
These errors I think you could have caught by running the local tests first 😉

jhawthorn:
😳 I think I'd lost a `#define` in a rebase 😅

maximecb:
Well I think you got it working this time.

Though I don't know why the Rubyspec and windows tests are "stuck". This is orthogonal to your PR, but any idea why that. might be happening?

PR #51, opened 2021-05-26
Add putstring to yjit codegen

jhawthorn:
Extracted from #49

This adds `putstring`, which is common in files which aren't using `frozen_string_literal: true`


maximecb:
Aaron thinks the cause could be this: https://github.com/Shopify/yjit/issues/54

maximecb:
I'll close this for now because we're blocked by #87  #54

PR #52, opened 2021-05-26
Increase default YJIT call threshold to 10. Add exec mem size arg.

maximecb:


maximecb:
@byroot changing the defaults make running on Shopify Core CI more seamless.

PR #57, opened 2021-05-28
Add checktype to yjit codegen

jhawthorn:
This implements the `checktype` instruction, which is used for string interpolation and array/hash destructuring (I'm mostly interested in string interpolation for now).

In the interpreter this is defined as `ret = (TYPE(val) == (int)type) ? Qtrue : Qfalse;`. I only implemented this for T_STRING, T_ARRAY, and T_HASH as these are the only types `compile.c` uses with checktype as far as I could tell. I also tried to use yjit's types to make this push a constant true/false when that could be determined (which I don't think will be too uncommon in string interp).

I also introduced a type for true/false, which I'll do more with in an upcoming PR.


yjit_codegen.c
```diff
@@ -1315,6 +1315,64 @@ gen_defined(jitstate_t* jit, ctx_t* ctx)
...
+        cmp(cb, REG0, imm_opnd(type_val));
+        mov(cb, REG1, imm_opnd(Qfalse));
+        cmovne(cb, REG0, REG1);
+
+        stack_ret = ctx_stack_push(ctx, TYPE_IMM);
+        mov(cb, stack_ret, REG0);
+
+        return YJIT_KEEP_COMPILING;
+    } else {
+        return YJIT_CANT_COMPILE;
```
maximecb, line 1372: Nitpick but I prefer exiting the function early if we can't compile :)

maximecb:
Looks pretty good 👌

maximecb:
Going to close this PR because #58 is a superset of this. Will merge #58.

PR #59, opened 2021-05-29
Detect true/false/nil/other types in YJIT branches when possible

jhawthorn:
Somewhat of a follow up to #57, where I added ETYPE_TRUE and ETYPE_FALSE

This adds checks on types to gen_branchnil/gen_branchif/gen_branchunless. Any time we don't have ETYPE_UNKNOWN we should have enough info to emit a direct jump.

This also adds checking for `Qtrue`/`Qfalse` in putobject and getinlinecache.

I saw the commented out `TEMP_CONST` in `enum yjit_temp_loc`, which this seems like it could be a fit for (and possibly avoid pushing true/false just to immediately pop it), but didn't dig much further.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -1904,17 +1904,30 @@ gen_branchif(jitstate_t* jit, ctx_t* ctx)
         yjit_check_ints(cb, side_exit);
     }
 
-    // Test if any bit (outside of the Qnil bit) is on
-    // RUBY_Qfalse  /* ...0000 0000 */
-    // RUBY_Qnil    /* ...0000 1000 */
-    x86opnd_t val_opnd = ctx_stack_pop(ctx, 1);
-    test(cb, val_opnd, imm_opnd(~Qnil));
-
     // Get the branch target instruction offsets
     uint32_t next_idx = jit_next_insn_idx(jit);
     uint32_t jump_idx = next_idx + jump_offset;
     blockid_t next_block = { jit->iseq, next_idx };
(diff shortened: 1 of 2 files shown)


yjit_codegen.c
```diff
@@ -1900,18 +1904,31 @@ gen_branchif(jitstate_t* jit, ctx_t* ctx)
...
     // Get the branch target instruction offsets
     uint32_t next_idx = jit_next_insn_idx(jit);
     uint32_t jump_idx = next_idx + jump_offset;
     blockid_t next_block = { jit->iseq, next_idx };
     blockid_t jump_block = { jit->iseq, jump_idx };
 
+    // Check if we know enough from type info
+    val_type_t val_type = ctx_get_opnd_type(ctx, OPND_STACK(0));
+    if (val_type.type != ETYPE_UNKNOWN) {
+        ctx_stack_pop(ctx, 1);
```
maximecb, line 1916: I think you would have to pop the value regardless of whether its type is known or not.

maximecb:
> I saw the commented out TEMP_CONST in enum yjit_temp_loc, which this seems like it could be a fit for (and possibly avoid pushing true/false just to immediately pop it), but didn't dig much further.

Yes, the idea is that for some values, we don't have to push anything.

I think I will hold off on merging this because I'm hoping with the new backend, we can merge comparisons and branches.

yjit_codegen.c
jhawthorn: There was a pop lower if we didn't hit this case, but they probably should have been combined for clarity.

PR #61, opened 2021-05-31
Use builtin_inline_p to skip a frame of C methods

k0kubun:
In CRuby, we're trying to annotate C methods that don't need a stack frame by `Primitive.attr! 'inline'`. Instead of hard-coding such a list of C methods in `cfunc_needs_frame`, this PR implements the same optimization using the annotation. 

If you use this annotation to implement the optimization, CI will automatically check if such a C method is indeed leaf. Note that `rb_hash_has_key` is one of the functions that do need a stack frame because it may call `rb_funcall`.

## Benchmark
```yml
# bench.yml
prelude: |
  def kernel_class
    self.class
  end
  10.times { kernel_class }
benchmark: kernel_class
loop_count: 50000000
```

```
$ benchmark-driver -v --rbenv 'before::yjit-before;after::yjit' bench.yml
before: ruby 3.1.0dev (2021-05-27T17:59:41Z main 5fa1c7f221) [x86_64-linux]
YJIT is enabled
after: ruby 3.1.0dev (2021-05-31T23:50:27Z inline-cfunc d6d6a9587a) [x86_64-linux]
YJIT is enabled
Calculating -------------------------------------
                         before       after
        kernel_class    33.942M     44.527M i/s -     50.000M times in 1.473080s 1.122914s

Comparison:
                     kernel_class
               after:  44527016.4 i/s
              before:  33942479.2 i/s - 1.31x  slower
```

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -2160,7 +2160,7 @@ gen_send_cfunc(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const 
     //print_ptr(cb, recv);
 
     // If this function needs a Ruby stack frame
-    const bool push_frame = cfunc_needs_frame(cfunc);
+    const bool push_frame = true;
 
     // Create a size-exit to fall back to the interpreter
     uint8_t *side_exit = yjit_side_exit(jit, ctx);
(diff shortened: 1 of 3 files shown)


maximecb:
I pushed this in #63 so I could fix the bugs.

PR #63, opened 2021-06-02
Use builtin_inline_p to avoid pushing a frame for primitive C methods

maximecb:


maximecb:
Thank you for the contribution k0kubun! :)

PR #68, opened 2021-06-04
Test CI #2

maximecb:


tool/lib/test/unit/parallel.rb
```diff
@@ -52,11 +52,14 @@ def _run_suite(suite, type) # :nodoc:
         # Thread for making sure the parallel worker is not stuck on one test
         # for too long.
         @timer_thread = Thread.new do
-          timeout = 8 * 60 # 8 minutes
+          timeout = 60 * 60 # 50 minutes
```
noahgibbs, line 55: This timeout doesn't match the comment.

PR #72, opened 2021-06-08
Implement gen_getlocal

jhawthorn:
This implements the general `getlocal` instruction based on the existing the generation code from `getlocal_wc1`, since this is the same just with more loops inside vm_get_ep.


maximecb:
Code looks good 👌 Out of curiosity. Why didn't you use `gen_getlocal_generic` for `getlocal_wc0`? Is there an inefficiency that would be introduced?

maximecb:
No you're right, the approach you took is best.

maximecb:
Up to +8% in railsbench now 😎

PR #81, opened 2021-06-10
add graphviz output

tenderlove:
This adds a method to blocks to get outgoing ids, then uses the outgoing
ids to generate a graphviz graph.

This is just a WIP / POC.  I wanted a more visual way to look at blocks and their relationships.  This commit makes a graphviz output so we can see which blocks point to which.  I think it would be interesting if we could

* highlight the jump addresses in some blocks
* Add better labels to the node shapes
* add information to the edges (maybe edge type?)

For this program:

```ruby
def threequals x
  case x
  when Array then 10
  when Hash then 10
  else
    10
  end
end

threequals([])
threequals([])
threequals([])
threequals([])
threequals({})
threequals({})
threequals({})
threequals({})
threequals({})

puts YJIT.graphviz_for(method(:threequals))

```

We get this output:

out

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -74,6 +74,7 @@ module YJIT
       # Subgraph contains disassembly
       buff << "  subgraph disasm {\n"
       buff << "  node [shape=record fontname=\"courier\"];\n"
+      buff << "  edge [fontname=\"courier\" penwidth=3];\n"
       blocks.each do |block|
         disasm = disasm_block(cs, block, highlight)
 
@@ -87,9 +88,9 @@ module YJIT
         buff << block.outgoing_ids.map { |id|
           next_block = blocks.bsearch { |nb| id <=> nb.id }
           if next_block.address == (block.address + block.code.length)
-            "b#{block.id} -> b#{id}[color=\"green\"];"
+            "b#{block.id} -> b#{id}[label=\"Fall\"];"
           else
-            "b#{block.id} -> b#{id};"
+            "b#{block.id} -> b#{id}[label=\"Jump\" style=dashed];"
           end
         }.join("\n")
         buff << "\n"


yjit_iface.c
```diff
@@ -1025,6 +1025,41 @@ rb_yjit_call_threshold(void)
...
+static VALUE
+block_id(VALUE self)
+{
+    block_t * block;
+    TypedData_Get_Struct(self, block_t, &yjit_block_type, block);
+    return PTR2NUM(block);
+}
+
+static VALUE
+outgoing_ids(VALUE self)
```
maximecb, line 1039: Should comment each function to explain what it's for, what it does.

maximecb:
Pretty cool! 

Small suggestions for improvements:
- Would be to render the graph with a monospace font.
- Comment above each new function/method explaining its purpose
- In terms of different types of edges, would be good to highlight fallthrough edges differently (ie: next block starts right after with no gap)

tenderlove:
> There are no jump instructions, and the addresses are very close. Is this a bug in my code, or do we have a problem with the code length?

(I will poke at this more tomorrow, as I'm sure I can figure it out.  It's just where I left off today 😄 )

jhawthorn:
Possibly related to this special case for getinlinecache? https://github.com/Shopify/yjit/blob/main/yjit_codegen.c#L438-L443

maximecb:
I like the font and styling of your graph 👍 

It's very possible that the end position of blocks is not getting correctly set/updated somewhere.

Normally we update the end position of blocks after rewriting fall-through branches: https://github.com/Shopify/yjit/blob/main/yjit_core.c#L596

I could use your help in debugging this, or possibly we could pair and look at it together tomorrow afternoon anytime after 11AM Eastern time. What do you say?

tenderlove:
out
OK! I made jumps dashed and falls solid.  I think this is ready to merge unless anyone has more feedback.

maximecb:
Looks good Aaron, thanks! :)

PR #82, opened 2021-06-11
Special-case !nil the way we already do with !true and !false.

noahgibbs:
With the yaml-load benchmark I'm working on for yjit-bench (https://github.com/Shopify/yjit-bench/pull/22) this reduces side-exits from opt_not to zero, where they were previously around 12% of side exits.

This runs correctly with "make test-all" for me locally.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -1975,7 +1975,7 @@ gen_opt_not(jitstate_t* jit, ctx_t* ctx)
     VALUE comptime_val = jit_peek_at_stack(jit, ctx, 0);
 
     // For the true/false case
-    if (comptime_val == Qtrue || comptime_val == Qfalse) {
+    if (comptime_val == Qtrue || comptime_val == Qfalse || comptime_val == Qnil) {
 
         // Get the operand from the stack
         x86opnd_t arg = ctx_stack_pop(ctx, 1);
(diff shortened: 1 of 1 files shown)


maximecb:
I think unfortunately you can't do that, because it's actually possible to define a `nil` method on objects!

noahgibbs:
Hm. This is testing for Qnil (the immediate nil value as an operand.) If you defined and called a nil method, wouldn't you get something other than immediate-nil-on-the-stack? I could be wrong, obviously.

This *does* seem to make a difference in my results with --yjit-stats, so I don't think I'm checking for something that never happens.

maximecb:
I think the code you wrote is correct, and it will result in less side-exits. However:
- If it sees Qnil at compilation time, it will side-exit some of the time when it encounters an object at run-time
- If it sees an object at compilation time, it will side-exit some of the time when it encounters nil at run-time

Ideally, we want the code to be as predictable as possible. We want as few side-exits as possible, and when there are side-exits, we also want those to be predictable. Right now, whether you side-exit or not is dependent on the run-time value you encounter.

noahgibbs:
Okay. So then we should skip this in favour of fixing the problem in a more general way later on?

This also sounds like the true/false case isn't working perfectly either, though I'm not seeing any side exits on opt_not in the benchmark runs I've seen.

This looks like adding the !nil runtime-only special case improves the speedup on yaml-load from 12% to 15% currently on the AWS instance.

maximecb:
> Okay. So then we should skip this in favour of fixing the problem in a more general way later on?

I'd kind of like to see a general fix if you're up for it. We can pair on it next week if you want, along with Aaron or Alan.

> This also sounds like the true/false case isn't working perfectly either, though I'm not seeing any side exits on opt_not in the benchmark runs I've seen.

The true/false case works pretty well in practice because people don't usually mix true/false with other types.

PR #83, opened 2021-06-11
Add warning with --yjit-stats in a Ruby not compiled with RUBY_DEBUG

noahgibbs:
I was talking to @tenderlove earlier and he mentioned that it's really annoying when you try to use --yjit-stats and it doesn't work because you didn't specify -DRUBY_DEBUG, or specified it like I did there without a nonzero numeric value. There's no message, it just silently does nothing.

Here's a PR that provides a warning when you specify --yjit-stats in a Ruby without -DRUBY_DEBUG=1.


maximecb:
This is a good idea, but it should probably error out rather than warn, and the logic should be placed in `rb_yjit_init` so that it work when the stats are enabled through an environment variable (see #84).

PR #84, opened 2021-06-11
Allow to enable `--yjit-stats` via a env variable

casperisfine:
### Context

I'd like to always print YJIT stats on our nightly CI, however it's a complex system and the vast majority of the code is shared with our regular ruby-head CI. 

So it's quite tricky to set `RUBYOPT="--yjit-stats"` only for YJIT builds, and even then it interfers with some scripts ran with regular ruby, which then abort complaining about that unknown option.

### Solution

This whole ordeal would be tremendously simpler if I could enable printing stats through an environment variable that would noop on regular ruby. This PR uses `YJIT_STATS=1`, but it could be anything.

### DIFF
--- a/yjit_iface.c
+++ b/yjit_iface.c
@@ -1035,6 +1035,8 @@ rb_yjit_init(struct rb_yjit_options *options)
     rb_yjit_opts = *options;
     rb_yjit_opts.yjit_enabled = true;
 
+    rb_yjit_opts.gen_stats |= !!getenv("YJIT_STATS");
+
     // Normalize command-line options to default values
     if (rb_yjit_opts.exec_mem_size < 1) {
         rb_yjit_opts.exec_mem_size = 256;


noahgibbs:
This looks like it would override a command-line option if it was given. I'll test it and see if I see that in practice, though.

noahgibbs:
Right now, not supplying YJIT_STATS means --yjit-stats doesn't work. I think it would make sense to print stats if either one is given.

maximecb:
Happy to hear you're working on getting YJIT to run nightly on the CI :)

I think Noah is right that in this current form, this breaks the command-line option.

casperisfine:
> Happy to hear you're working on getting YJIT to run nightly on the CI :)

It already does. I just don't get the stats out /

> I think Noah is right that in this current form, this breaks the command-line option.

Yeah, I just did this PR quick & dirty, I'll try to figure what's wrong and fix it.

Are you ok with the general principle?

casperisfine:
> I think it would make sense to print stats if either one is given.

That was the intent yeah. 

I saw:

```c
    else if (opt_match_noarg(s, l, "stats")) {
        yjit_opt->gen_stats = true;
    }
``

So I assumed initializing it with the env var would work. I'll dig deeper.

casperisfine:
Ok, I was indeed confused as to the order of execution. Should be good to go now.

PR #86, opened 2021-06-15
Allow yjit.rb to work with frozen strings

jhawthorn:
This will be necessary if https://github.com/ruby/ruby/pull/4573 lands (and otherwise shouldn't hurt anything)

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -6,7 +6,7 @@ module YJIT
       blocks = YJIT.blocks_for(iseq)
       return if blocks.empty?
 
-      str = ""
+      str = String.new
       str << iseq.disasm
       str << "\n"
 


maximecb:
Happy to merge this but can you explain...

By default string literals "" are not frozen, but `String.new` is?

jhawthorn:
Yeah! When the `frozen_string_literals` compile option is enabled (it isn't by default, but it's very common to enable it per-file, and I think it should be enabled for the builtin `.rb`s) string literals like `""` are frozen. So to get a non-frozen string to use as our buffer we must can `String.new` or `"".dup` (or a few other options).

maximecb:
Thanks John :)

PR #90, opened 2021-06-16
Improve opt_not by expanding cfunc codegen

XrXr:
Two commits. Resulting code for `!obj` where `obj` is a heap object:

```
  ; opt_not
  111468913:  mov	rax, qword ptr [rdx]
  ; guard not immediate
  111468916:  test	al, 7
  111468919:  jne	0x119468160
  11146891f:  cmp	rax, 8
  111468923:  jbe	0x119468160
  ; guard known class
  111468929:  movabs	rcx, 0x10dffe7f0
  111468933:  cmp	qword ptr [rax + 8], rcx
  111468937:  jne	0x119468188
  ; rb_obj_not(truthy)
  11146893d:  mov	qword ptr [rdx], 0
  111468944:  jmp	0x1114688dd
```

Output when `obj == nil`:

```
  ; opt_not
  111468867:  mov	rax, qword ptr [rdx]
  ; guard object is nil
  11146886a:  cmp	rax, 8
  11146886e:  jne	0x1114688b1
  ; rb_obj_not(nil_or_false)
  111468874:  mov	qword ptr [rdx], 0x14
```

Output when `obj == false` is similar.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -1962,51 +1962,8 @@ gen_opt_str_uminus(jitstate_t* jit, ctx_t* ctx)
 }
 
 static codegen_status_t
-gen_opt_not(jitstate_t* jit, ctx_t* ctx)
+gen_opt_not(jitstate_t *jit, ctx_t *ctx)
 {
-    // Defer compilation so we can specialize type of argument
-    if (!jit_at_current_insn(jit)) {
-        defer_compilation(jit->block, jit->insn_idx, ctx);
-        return YJIT_END_BLOCK;
-    }
-
-    uint8_t* side_exit = yjit_side_exit(jit, ctx);
-
-    VALUE comptime_val = jit_peek_at_stack(jit, ctx, 0);
-
-    // For the true/false case
-    if (comptime_val == Qtrue || comptime_val == Qfalse) {
-
-        // Get the operand from the stack
-        x86opnd_t arg = ctx_stack_pop(ctx, 1);
-
-        uint32_t DONE = cb_new_label(cb, "DONE");
-
-        // Qtrue => Qfalse
-        mov(cb, REG0, imm_opnd(Qfalse));
-        cmp(cb, arg, imm_opnd(Qtrue));
-        je_label(cb, DONE);
-
-        // Qfalse => Qtrue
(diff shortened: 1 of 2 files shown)


yjit_codegen.c
```diff
@@ -2268,6 +2253,48 @@ jit_protected_callee_ancestry_guard(jitstate_t *jit, codeblock_t *cb, const rb_c
...
+    return true;
+}
+
+// Check if we know how to codegen for a particular cfunc method
+static cfunc_codegen_t
+lookup_cfunc_codegen(const rb_method_cfunc_t *cfunc)
+{
+    if (cfunc->func == rb_obj_not) {
+        return jit_rb_obj_not;
+    }
```
jhawthorn, line 2294: If we adopt this pattern I have [a bunch more cfuncs](https://github.com/Shopify/yjit/compare/main...jhawthorn:specialized_methods#diff-746b53169139d506c1263176830553fb45c3a581c032e337f432bebb4754e2e7R2194) to add here 😁

yjit_codegen.c
```diff
@@ -2268,6 +2253,48 @@ jit_protected_callee_ancestry_guard(jitstate_t *jit, codeblock_t *cb, const rb_c
...
+    if (recv_opnd.type == ETYPE_NIL || recv_opnd.type == ETYPE_FALSE) {
+        ADD_COMMENT(cb, "rb_obj_not(nil_or_false)");
+        mov(cb, out_opnd, imm_opnd(Qtrue));
+        ctx_set_opnd_type(ctx, OPND_STACK(0), TYPE_TRUE);
+    }
+    else if (recv_opnd.is_heap || recv_opnd.type != ETYPE_UNKNOWN) {
+        // Note: recv_opnd.type != ETYPE_NIL && recv_opnd.type != ETYPE_FALSE.
+        ADD_COMMENT(cb, "rb_obj_not(truthy)");
+        mov(cb, out_opnd, imm_opnd(Qfalse));
+        ctx_set_opnd_type(ctx, OPND_STACK(0), TYPE_FALSE);
```
maximecb, line 2277: IMO this would look cleaner if you did a `return true` in the if statements and avoided the `else` and `return true` at the end.

PR #92, opened 2021-06-18
Use an st_table for optimized method codegen

jhawthorn:
We recently added the ability to optimize a known cfunc with custom code generation for it. I think this is going to be super helpful for yjit and probably more convenient in many cases than tracking BOPs (basic operations).

Previously we performed this lookup with a switch statement on the address of the c function being called. This commit swaps that out for a hash lookup on the method definition. For now I've kept this limited this to cfuncs, but it wouldn't take significant changes to make this work for other method types (we would need to init the table later in the boot sequence to pickup builtins though).

This implementation is similar to how the interpreter keeps track of which BOPs are redefined

This has a few advantages:
- Doesn't the C function's symbol to be exported (they're often static)
- This could support `VM_METHOD_TYPE_OPTIMIZED` in the future.
- This could support `VM_METHOD_TYPE_ISEQ` in the future. Kernel#class would be a good candidate for this since to yjit it will just be a constant push as we already know the class through BBV.
- Slightly cleaner to declare
- Less tightly coupled to each method's implementation

And a couple minor trade-offs:
- The looser coupling could be seen as a disadvantage (I don't think so, but could appreciate this perspective)
- If a cfunc is defined multiple times we would need to declare it on each definition. ex. `BasicObject#==` and `BasicObject#equal?`. This is rare compared to using an alias.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -19,6 +19,9 @@
 // Map from YARV opcodes to code generation functions
 static codegen_fn gen_fns[VM_INSTRUCTION_SIZE] = { NULL };
 
+// Map from method entries to code generation functions
+static st_table *yjit_method_codegen_table = NULL;
+
 // Code block into which we write machine code
 static codeblock_t block;
 codeblock_t* cb = NULL;
@@ -2270,7 +2273,7 @@ jit_protected_callee_ancestry_guard(jitstate_t *jit, codeblock_t *cb, const rb_c
 }
 
 // Return true when the codegen function generates code.
-typedef bool (*cfunc_codegen_t)(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const rb_callable_method_entry_t *cme, rb_iseq_t *block, const int32_t argc);
+typedef bool (*method_codegen_t)(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const rb_callable_method_entry_t *cme, rb_iseq_t *block, const int32_t argc);
 
 // Codegen for rb_obj_not().
 // Note, caller is responsible for generating all the right guards, including
@@ -2302,11 +2305,12 @@ jit_rb_obj_not(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const 
 }
 
 // Check if we know how to codegen for a particular cfunc method
-static cfunc_codegen_t
-lookup_cfunc_codegen(const rb_method_cfunc_t *cfunc)
+static method_codegen_t
+lookup_cfunc_codegen(const rb_method_definition_t *def)
 {
-    if (cfunc->func == rb_obj_not) {
-        return jit_rb_obj_not;
+    method_codegen_t gen_fn;
+    if (st_lookup(yjit_method_codegen_table, def->method_serial, (st_data_t *)&gen_fn)) {
+        return gen_fn;
     }
     return NULL;
 }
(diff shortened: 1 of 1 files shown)


XrXr:
I think I'm negative on this. The `lookup_cfunc_codegen` setup is for really simple C methods, and basically YJIT's equivalent for the interpreter's setup with using `vm_method_cfunc_is` and substituting.

> Doesn't the C function's symbol to be exported (they're often static)

I think we only want to do it for ones that already go through `vm_method_cfunc_is()`, which are already visible in yjit's units. By the way, Ruby builds with `-fvisibility=hidden` so non-static functions are not exported by default.

> Less tightly coupled to each method's implementation

That's a negative in my eye. If one changes the other should too. The coupling is inherent to re-implementing. Hiding it might hide bugs.

By the way this substitution comes with a caveat in that it doesn't check for interrupts. For the `opt_not` case this is okay, because the interpreter doesn't check it either. This is one of the reasons why I think we want to stick to substituting very simple methods that the interpreter already does.

jhawthorn:
> The `lookup_cfunc_codegen` setup is for really simple C methods, and basically YJIT's equivalent for the interpreter's setup with using `vm_method_cfunc_is` and substituting.

We're more or less on the same page here, but I want this to be YJIT's equivalent to both `vm_method_cfunc_is` **and BOPs**.

IMO these two concepts are the same. In the interpreter BOPs are used when the class is known (ex. `nil.nil?` AKA `BOP_NIL_P`) and `vm_method_cfunc_is` is used when we want to optimize on any class `anything_else.nil?` (`vm_method_cfunc_is(iseq, cd, recv, rb_false)`). I think the only reason they are different is that the interpreter wanted a fixed number of classes to efficiently check BOPs, and wanted to be able to use inline method caches in the other cases. Neither of these apply to our JIT.

As I mentioned this implementation is very similar to how [the BOP method table](https://github.com/ruby/ruby/blob/master/vm.c#L1881-L1889) is initialized. So I think it makes sense to do similar here.

_sidenote: currently BOPs also only allow cfuncs, but I think that's an artificial limitation and could easily see (👀 `NilClass#nil?`) implemented in Ruby builtins in the future_

> > Less tightly coupled to each method's implementation
> 
> That's a negative in my eye. If one changes the other should too. The coupling is inherent to re-implementing. Hiding it might hide bugs.

I agree that in an ideal world we'd be alerted of (or at least read) any changes to the methods we're overriding (why I listed it as a trade off). However I think this approach vs the switch statement makes a tiny difference in the kinds of changes that would error for us, I can only think of renaming the cfunc or converting the cfunc to an iseq builtin. FWIW the BOPs would "miss" any changes this PR would "miss". At the end of the day we have to rely on tests 😅.

> By the way this substitution comes with a caveat in that it doesn't check for interrupts. For the `opt_not` case this is okay, because the interpreter doesn't check it either. This is one of the reasons why I think we want to stick to substituting very simple methods that the interpreter already does.

I disagree with this. We're clearly okay with dropping checks for interrupts both in YJIT and in upstream. In YJIT we don't check for _forward_ branches anymore. Upstream it's been fine to mark builtin calls with `Primitive.attr! 'inline'`, and `opt_` instructions aren't concerned with this.

---

In any case we could use this PR's implementation and decide what is appropriate later (when there are concerte examples) of:
- only `cfunc_is`
- ☝️ + a few select BOPs
- ☝️ + all/most BOPs
- ☝️ + a few more methods which make sense in YJIT that wouldn't fit as well in the interpreter (happy to discuss next week)

maximecb:
You bring up some good points Alan.

I see your point wrt static but I think at the same time it is easier not to directly deal with the C method symbols because that reduces the YJIT footprint in terms of changes required to merge upstream. Having to export everything as non-static means we have to make lots of small changes in many source files 🤔 

With respect to dropping interrupt checks, I think this is mostly a problem in places where we could have loops. We don't want to be stuck in a loop for a long time and not check for interrupts. For small methods it's probably ok. In an ideal world, we would have some kind of estimate how many instructions ran since the last interrupt check and use that, eg: make sure we do an interrupt check every 10K bytecode instructions 🤔

maximecb:
@jhawthorn if you fold `yjit_init_optimized_methods` into `yjit_init_codegen` for consistency we will merge this PR 👍 

However, we'd like to stick to only reimplement small trivial methods in the JIT for now, because we don't want to end up reimplementing large chunks of C or Ruby code in the JIT if possible, as that seems error-prone. And for Ruby code, eventually, we will hopefully have inlining.

jhawthorn:
> if you fold `yjit_init_optimized_methods` into `yjit_init_codegen` for consistency we will merge this PR

Done.

In the future we might want to pull it back out, as builtins are run later and we may want to override those, but since this currently only aims to support cfuncs we can decide on that later.

PR #94, opened 2021-06-21
Remove some MicroJIT vestiges

tenderlove:
Just happened to run across this, so lets fix them so we can stay On Brand


noahgibbs:
LGTM

maximecb:
😮

PR #95, opened 2021-06-21
Disable invokesuper codegen for now. Add test

XrXr:
The added test fails with SystemStackError with --yjit-call-threshold=1.

---

I haven't had time to dig into the details for this, but in general I think `invokesuper` doesn't fit into YJIT's current invalidation scheme nicely. The way it uses `assume_method_lookup_stable()` works, but is stepping into territory that the function wasn't designed to do. I wouldn't be surprised if there are other problems.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1204,3 +1204,23 @@ assert_equal '123', %q{
   foo(Foo)
   foo(Foo)
 }
+
+# invokesuper edge case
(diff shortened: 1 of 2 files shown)


jhawthorn:
Good find. I'd guess this might be related to [this check](https://github.com/Shopify/yjit/blob/0f9d587b1f157f6406840ebbee5f106047ec2731/yjit_codegen.c#L3030-L3035) not being sufficient to prevent issues. Happy to look into this further

PR #97, opened 2021-06-22
Fix incorrect recursion in invokesuper

jhawthorn:
A potential fix for #95

This adjusts the ancestor check to check not for just re-including the same module but any case where an ancestor has the same method. This fixes Alan's test case (and the existing Ruby tests for similar behaviours), but I'm not 100% sure there isn't still some edge case it won't catch (requires more thought).

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -3050,7 +3050,7 @@ gen_invokesuper(jitstate_t *jit, ctx_t *ctx)
     const rb_callable_method_entry_t *ancestor_cme;
     while (ancestor) {
         ancestor_cme = rb_callable_method_entry(ancestor, mid);
-        if (ancestor_cme && ancestor_cme->def == cme->def) {
+        if (ancestor_cme && ancestor_cme->def == me->def) {
             return YJIT_CANT_COMPILE;
         }
         ancestor = RCLASS_SUPER(ancestor);
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -2972,7 +2972,6 @@ gen_send(jitstate_t *jit, ctx_t *ctx)
     return gen_send_general(jit, ctx, cd, block);
 }
 
-RBIMPL_ATTR_MAYBE_UNUSED() // not in use as it has problems in some situations.
 static codegen_status_t
 gen_invokesuper(jitstate_t *jit, ctx_t *ctx)
 {
@@ -3027,13 +3026,6 @@ gen_invokesuper(jitstate_t *jit, ctx_t *ctx)
         return YJIT_CANT_COMPILE;
     }
 
-    // Because we're assuming only one current_defined_class for a given
-    // receiver class we need to check that the superclass doesn't also
-    // re-include the same module.
-    if (rb_class_search_ancestor(comptime_superclass, current_defined_class)) {
-        return YJIT_CANT_COMPILE;
-    }
-
     // Do method lookup
     const rb_callable_method_entry_t *cme = rb_callable_method_entry(comptime_superclass, mid);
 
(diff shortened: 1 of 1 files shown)


XrXr:
Looks like this makes it compare `cme` against itself and bails unconditionally:

```ruby
class A
  def itself
    super
  end
end

A.new.itself
puts YJIT.disasm(A.instance_method(:itself))
```

```
== disasm: #<ISeq:itself@test.rb:2 (2,2)-(4,5)> (catch: FALSE)
0000 putself                                                          (   3)[LiCa]
0001 invokesuper                            <calldata!mid:itself, argc:0, FCALL|ARGS_SIMPLE|SUPER|ZSUPER>, nil
0004 leave                                                            (   4)[Re]

== BLOCK 1/2: 7 BYTES, ISEQ RANGE [0,4) ========================================
  ; putself
  10f66885c:  mov	rax, qword ptr [rdi + 0x18]
  10f668860:  mov	qword ptr [rdx], rax
== BLOCK 2/2: 40 BYTES, ISEQ RANGE [1,1) =======================================
  ; invokesuper
  ; exit to interpreter
  10f668863:  lea	rdx, [rdx + 8]
  10f668867:  mov	qword ptr [rdi + 8], rdx
  10f66886b:  mov	qword ptr [rsi + 0x10], rdi
  10f66886f:  movabs	rax, 0x7f86fd92e218
  10f668879:  mov	qword ptr [rdi], rax
  10f66887c:  mov	rcx, qword ptr [rbx + 0x10]
  10f668880:  mov	rax, qword ptr [rcx]
  10f668883:  mov	r13, rax
  10f668886:  mov	r8, rcx
  10f668889:  jmp	qword ptr [rax]
```

Unrelated, here is a test case that might be interesting:

```ruby
# threshold = 1
module M
  def foo = :m
end

class B
  include M
  def foo = super
  alias bar foo
  remove_method :foo
end

p B.new.bar

module M
  def foo = :mredef
end

p B.new.bar
```

maximecb:
It looks like, to test things more rigorously, we really need some way to assert that a method has never side-exited. Maybe we could store a one-bit flag on iseqs that we set when in debug mode? WDYT?

noahgibbs:
It could definitely be interesting to have more per-method statistics. I'm not sure how many of them should be RUBY_DEBUG-only. But it would be nice to assert things like "this method never side exits" or "this method has never been de-optimised" or "this method has no blocks treated as megamorphic."

XrXr (changes requested):
This change is not enough to fix the problem. The following gives the wrong answer:
```ruby
# --yjit-call-threshold=1
module M
  def foo = :m
end

class B
  include M
  def foo = [:B, super]
end

ins = B.new
ins.singleton_class # materialize the singleton class
p ins.foo

ins.singleton_class.define_method(:bar, B.instance_method(:foo))
p ins.bar
```

https://github.com/Shopify/yjit/blob/9a3c7ea748a1b5cbaba5593b67bfe92a455f1755/yjit_codegen.c#L3085
`gen_invokesuper()` uses `assume_method_lookup_stable()` in a way that it isn't designed for.
`assume_method_lookup_stable()`'s comment explains that it gurantees that `rb_callable_method_entry(receiver_klass, mid) == cme`.
That equality doesn't hold at the time `gen_invokesuper` calls it. `cme` is from `comptime_superclass`.
`assume_method_lookup_stable()`'s guarantee about `cme` being valid happens to catch some scenarios where
the output code ought to be invalidated, but it's incomplete.

The problem boils down to the current runtime guards and implicit checks not being sufficient to ensure that
`RCLASS_SUPER(CLASS_OF(recv))` is the consistent between runtime and compile time. I think `invokesuper`
inherently needs special handling in the JIT for correctness. Passing existing tests isn't enough. By nature,
optimized output code has more ways to be wrong than the interpreter.

There might be other bugs I haven't thought of. Maybe some involving refinements, maybe some involving that `BMETHOD` check at compile time. It needs a lot of special care to get right. To help limit liability, I think `gen_invokesuper` should stay disabled. I will add an assert to `assume_method_lookup_stable()` and this test in a separate PR.

Issue #98, opened 2021-06-22
Construct to assert that a method never exited to the interpreter


maximecb:
It looks like we could really use a way to verify that Ruby methods never abnormally exit to the interpreter. This would help us write more rigorous tests. At the moment, we semi-frequently run into cases where we write test code, and it appears to behave correctly, but in fact, we're side-exiting to the interpreter early, and never running the code we want to test in the JIT. This is a silent failure.

We already store an array of blocks per iseq. I think that we could also store a one-bit flag for each iseq. In debug mode, when taking an "abnormal" side-exit, we would set this flag to 1.

One caveat here is that there are legitimate reasons to side-exit that are not a "failure" on the JIT's part. For example, when we do interrupt checks, we infrequently exit to the interpreter to handle those. AFAIK, during tests, provided there are no ractors and we don't allocate anything, this should not happen, so we should still be able to write functions in a way that we know they won't side-exit, but @XrXr your opinion is welcome.

Syntax idea (open to alternatives too):
```
YJIT.never_exited(method(:foo))
```

@jhawthorn I know this is less fun than codegen but this could be a good issue for you to work on if you're open to it. We have to be very rigorous when testing, otherwise we will all die a segfaulty death.

jhawthorn:
> We already store an array of blocks per iseq. I think that we could also store a one-bit flag for each iseq. In debug mode, when taking an "abnormal" side-exit, we would set this flag to 1.

Alternatively could we use the existing (global) side exit counts?

maximecb:
That seems somewhat less clean to me, because we'd have to either keep track of the old count from Ruby, or always reset it to zero. We're currently getting stats from the Shopify Core CI, so if we reset the count to zero, those stats would suddenly be no longer valid. So that leaves getting the count before, and the count after, as the other option. Maybe there's a way to wrap that in a Ruby block or something? I guess that's doable?

jhawthorn:
Yeah, I think a Ruby block could make that ergonomic, we've done similar to record the difference in GC stats.

One thing I like about using the global side exist counts is that we can differentiate between the different types of counted side exits.

I was thinking more along the lines of adding a test similar to this one from MJIT https://github.com/ruby/ruby/blob/master/test/ruby/test_jit.rb which runs given code and has assertions on what was compiled, etc. I'll try to find some time to attempt this.

maximecb:
Sounds good. If you want to sketch what the API could look and comment in this issue I'd be curious what you're thinking.

[...]

PR #103, opened 2021-06-24
Add test of YJIT compilation

jhawthorn:
Addresses (at least part of) #98

This takes a similar approach to an existing test for MJIT, which compiles small snippets and asserts that the instruction intended to be tested actually existed in the script given. This test goes a step further and asserts that YJIT ended up compiling at least one block for the specific instruction we're testing. This also optionally performs an assert on the expected number of side exits per-opcode.

To do this it invokes Ruby in a subprocess, so that we have full control over the runtime (🤞 no interrupts) and can reset our stats without worry. Runtime stats are sent back Marshal'd over a file descriptor.

I think this is maybe best illustrated by looking at some tests which would fail (which I legitimately ran into when writing a few samples 😅)

---

``` ruby
assert_compiles('"" == ""', insns: %i[opt_eq], stdout: 'true')
```

```
TestYJIT#test_compile_eq_string [/home/jhawthorn/src/yjit/test/ruby/test_yjit.rb:32]:
Expected to compile instructions opt_eq but didn't.
Compiled ranges: [0...0]
iseq:
== disasm: #<ISeq:<compiled>@<compiled>:1 (1,0)-(1,8)> (catch: FALSE)
0000 putstring                              ""                        (   1)[Li]
0002 putstring                              ""
0004 opt_eq                                 <calldata!mid:==, argc:1, ARGS_SIMPLE>
0006 leave
```

Fails because we weren't able to compile `putstring` and the `opt_eq` never ended up in any blocks. You can see the working version of this test included in this commit which uses `-""` (`opt_uminus`) instead.

---

``` ruby
assert_compiles('123.nil?', insns: %i[opt_nil_p], stdout: 'false')
```

```
TestYJIT#test_compile_opt_nil_p [/home/jhawthorn/src/yjit/test/ruby/test_yjit.rb:24]:
Expected no exits, but got
{:opt_nil_p=>1}
```

You can see the working version of this test included, which specifies that we're allowed to side exit once on this instruction (but that it still needs to be compiled)

---

**TODO**
- [x] ~Currently this puts the given script into a method. We probably want to evaluate at the top level and allow defining our own methods, but that makes finding the right iseqs (the mjit tests look for iseqs recursively) and matching them up with blocks a little more difficult.~ Fixed, now recursively checks iseqs in the provided script
- [x] ~I'm not certain the way I'm tracking iseq offsets is correct~ seems good
- [x] ~I'm not sure if side exits due to interrupts are a concern~ hasn't been a problem
- [x] ~When the scripts errors, the output is a bit messy due to being run with `--yjit-stats`. Maybe we want to separate recording stats from printing stats?~ Might as well do this in a separate PR
- [x] Need to skip this test if compiling without RUBY_DEBUG
- [x] Need to skip this test on unsupported platforms

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -8680,6 +8680,18 @@ mjit_build_dir.$(OBJEXT): {$(VPATH)}internal/compiler_since.h
 mjit_build_dir.$(OBJEXT): {$(VPATH)}internal/config.h
 mjit_build_dir.$(OBJEXT): {$(VPATH)}ruby-runner.c
 mjit_build_dir.$(OBJEXT): {$(VPATH)}ruby-runner.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}config.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/apple.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/clang.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/gcc.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/intel.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/msvc.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_is/sunpro.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/compiler_since.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}internal/config.h
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}ruby-runner.c
+mjit_build_dir.so-ruby-runner.$(OBJEXT): {$(VPATH)}ruby-runner.h
 mjit_compile.$(OBJEXT): $(CCAN_DIR)/check_type/check_type.h
 mjit_compile.$(OBJEXT): $(CCAN_DIR)/container_of/container_of.h
 mjit_compile.$(OBJEXT): $(CCAN_DIR)/list/list.h
@@ -17202,6 +17214,7 @@ yjit_iface.$(OBJEXT): {$(VPATH)}vm_debug.h
 yjit_iface.$(OBJEXT): {$(VPATH)}vm_opts.h
 yjit_iface.$(OBJEXT): {$(VPATH)}vm_sync.h
 yjit_iface.$(OBJEXT): {$(VPATH)}yjit.h
+yjit_iface.$(OBJEXT): {$(VPATH)}yjit.rb
 yjit_iface.$(OBJEXT): {$(VPATH)}yjit.rbinc
 yjit_iface.$(OBJEXT): {$(VPATH)}yjit_asm.h
 yjit_iface.$(OBJEXT): {$(VPATH)}yjit_codegen.h
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -9,54 +9,54 @@ return unless YJIT.enabled?
 # insipired by the MJIT tests in test/ruby/test_jit.rb
 class TestYJIT < Test::Unit::TestCase
   def test_compile_putnil
-    assert_compiles('nil', insns: %i[putnil], stdout: 'nil')
+    assert_compiles('nil', insns: %i[putnil], result: nil)
   end
 
   def test_compile_putobject
-    assert_compiles('true', insns: %i[putobject], stdout: 'true')
-    assert_compiles('123', insns: %i[putobject], stdout: '123')
-    assert_compiles(':foo', insns: %i[putobject], stdout: ':foo')
+    assert_compiles('true', insns: %i[putobject], result: true)
+    assert_compiles('123', insns: %i[putobject], result: 123)
+    assert_compiles(':foo', insns: %i[putobject], result: :foo)
   end
 
   def test_compile_opt_not
-    assert_compiles('!false', insns: %i[opt_not], stdout: 'true')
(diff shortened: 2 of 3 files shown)


yjit_iface.c
```diff
@@ -707,7 +707,31 @@ comments_for(rb_execution_context_t *ec, VALUE self, VALUE start_address, VALUE
     return comment_array;
 }
 
+// Primitive called in yjit.rb. Export all exit counters as a Ruby hash.
+static VALUE
+get_exit_counters(rb_execution_context_t *ec, VALUE self)
+{
+#if RUBY_DEBUG
 // Primitive called in yjit.rb. Export all runtime counters as a Ruby hash.
```
noahgibbs, line 715: Minor nitpick: I think this comment line is now a repeat.

noahgibbs:
I agree that we'll want some way to collect stats without printing them. I'm noticing the same thing in some YJIT CI work.

noahgibbs:
Not sure if we want to hold this one up until we finalise our preferred interface to exit_op_count. This PR uses its own, exposing it via a separate method called YJIT.runtime_exits. It would be easy to convert the code later if we wanted to, or we could finalise that interface and then modify this PR.

Other than that, this LGTM. @maximecb, you may want to check the use of heredocs (Ruby multiline strings). They look appropriate and tasteful to me but I know they're not always your thing. He's using the indented style, which may help.

maximecb:
> Not sure if we want to hold this one up until we finalise our preferred interface to exit_op_count.

@noahgibbs Yes we probably want to wait a little bit to merge. One thing that you should probably copy for your PR though is to make sure to take the VM and ractor locks before producing the counters. Otherwise it's not safe to return the counters before the execution is over.

@jhawthorn You've added a number of entries to `common.mk` that are related to MJIT. Is that coming from the upstream `ruby/ruby` repo?

> To do this it invokes Ruby in a subprocess, so that we have full control over the runtime (🤞 no interrupts) and can reset our stats without worry. Runtime stats are sent back Marshal'd over a file descriptor.

How fast is that? I'm assuming the individual tests don't run in parallel, but the file ` test/ruby/test_yjit.rb` is run in parallel with other tests in `test-all` so it doesn't matter much? In order to help prevent interrupts, maybe we could trigger a full GC before runing the test, if that makes sense?

My main feedback is that the code needs readability improvements and commenting. It's seem scomplex and hard to parse to me, and maybe it's trying to do to much at once, because you have one call `assert_compiles` that's actually trying to do 4 things:

- Verify that the script compiles
- Check that specific instructions have been compiled
- Check that the standard output is some specific string
- Verify that the exit counts are equal to given values

You can also manually specify a `min_calls` which is going to change the behavior.

Probably, we always want to run those tests with `call_threshold` 1 for simplicity. Otherwise, I think that `min_calls` should be renamed to `call_threshold` for consistency (otherwise people will have to look up what it's used for), and this value should only specified if the value is not 1. 

The use of heredoc makes this pretty confusing IMO. I wish there was a way to have unnamed multiline strings, but if you're going to use heredocs, they should probably be the last argument to the `assert_compiles` method... Otherwise having the other arguments interleaved with it makes it hard to read (at least for me).

It may also be better if we just had `assert_no_exits` method instead of the complex `assert_compiles` form. That would be simpler and more intuitive. I think it's what we want most of the time?

maximecb:
This will need to be updated following the changes Noah made to our counter handling.

noahgibbs:
Hey, John.

With the YJIT PR to include exit reasons in YJIT stats, I was wondering if this could be reasonably simplified to not use a subprocess. It looks like a min_calls setting higher than 1 is just being used to test min_calls itself, which probably doesn't need to happen separately for every instruction. And it seems like exit-reasons and instructions-in-disassembly can now be tested without having to invoke a Ruby subprocess.

That wouldn't check stdout, but (it looks like to me) stdout seems to only be verifying the correctness of the JITted Ruby code -- we'd normally do that via "make test-all" and its relatives. For instance, I don't think we need to verify the correctness of string interpolation at the same time we make sure string interpolation generates code using checktype and concatstrings.

I'm also not sure if I'm missing some functionality here - let me know if I'm not seeing things that get verified that would need a subprocess.

But if this *can* be done with no subprocess, that will probably reduce the complexity and runtime of the tests.

jhawthorn:
I am out of office this week and part of next, so won't be taking a close look until then.

> It looks like a min_calls setting higher than 1 is just being used to test min_calls itself, which probably doesn't need to happen separately for every instruction

There are a few things we would need/want `min_calls` for (though we might want to name it `call_threshold`, I used the same naming as the [existing MJIT tests](https://github.com/ruby/ruby/blob/master/test/ruby/test_jit.rb#L1164)). For a start, `opt_getinlinecache` can only be compiled if it's been run through the interpreter previously (to set the inline cache). It may also be helpful to be able to test non-jitted code calling into jitted code and vise versa. Finally, yeah, we might want to test the call-threshold calculation itself. The default is `1`, which is usually the most convenient and I think there's a big advantage to having it be configurable and consistent (most of the tests in `bootstraptest` no longer work with the default arguments because we wrote them assuming `call_threshold=2` and the default is now `10`).

> stdout seems to only be verifying the correctness of the JITted Ruby code -- we'd normally do that via "make test-all" and its relatives. For instance, I don't think we need to verify the correctness of string interpolation at the same time we make sure string interpolation generates code using checktype and concatstrings

I agree. Checking stdout is optional (if not specified in the test it's not checked) but is quite useful to check that the test code provided does what is expected. The [existing similar test for MJIT also checks stdout](https://github.com/ruby/ruby/blob/master/test/ruby/test_jit.rb).

> But if this _can_ be done with no subprocess, that will probably reduce the complexity and runtime of the tests.

Runtime, definitely, but complexity I'm not sure sure. Having a subprocess is very reliable (I don't know if this would be true using the same process) and makes some tests much easier (especially with control of call-threshold). I also think it's really helpful to mirror the behaviour of the [existing MJIT tests](https://github.com/ruby/ruby/blob/master/test/ruby/test_jit.rb#L1164).

I could see us including separate test files with both approaches.

jhawthorn:
> it would be nice to have the script as the last argument

If we want to accept kwargs, they have to come after the positional argument (though I suppose we could make `src:` a keyword argument, I like that this still somewhat matches `test/ruby/test_jit.rb`)

I made some updates: the tests now check a `result:` instead of `stdout:` which is a bit cleaner (no need to match inspect format exactly, better errors, less typing).

Updated this to the new statistics from #111 and also updated the tests as we have fewer side exits now 😁.

maximecb (approved):

maximecb:
Looks good 👍

noahgibbs (approved):

PR #105, opened 2021-06-24
Add a macro for tweaking default call threshold

XrXr:
For use in development for writing `test.rb` and running with `make
run`.


noahgibbs:
LGTM

PR #108, opened 2021-06-28
Remove seven YJIT counters that are no longer used.

noahgibbs:


noahgibbs:
Not sure what's up with the check that's failing. It shows a download failure, so I'm re-queueing it to see if it fails consistently.

noahgibbs:
Yeah, I'm getting "failed to download Github Action" and Kubernetes "ContainerID cannot be null" errors. I'll wait and re-queue.

PR #110, opened 2021-06-28
Partial revert of ceebc7fc98d

tenderlove:
I'm looking through the places where YJIT needs notifications.  It looks
like these changes to gc.c and vm_callinfo.h have become unnecessary
since 84ab77ba592.  This commit just makes the diff against upstream
smaller, but otherwise shouldn't change any behavior.


XrXr (approved):
Thanks for whittling down the diff!

Issue #107, opened 2021-06-28
Refactoring `get_stat_counters`


maximecb:
Noah would like to refactor `get_stat_counters` so it exposes more stats to Ruby code, which he is going to use for the `yjit-metrics` project. The main problem we have is that the exit reasons are mixed with many other counters. I agree with the general idea that our stats/counters probably should be better organized. My main concern right now is that the C code is a bit messy, and I would like the C code to get cleaner, not messier.

I think it would make sense having `get_stat_counters` return a struct which has different fields in it. AFAIK, semantically all the values we have so are are "counters". They're all integer values that get incremented. So I would avoid referring to some of them as counters and others not.

AFAIK exit_op_count are just "toplevel" counts of how many times we exit for a given instruction. Whereas the yjit_runtime_counters are assorted counters, which include a breakdown of "exit reasons" for the `exit_op_count`. 

It would be kind of nice if the C code could separate the exit reason counters into their own hash. That seems cleaner, though that could be annoying to do in C because we're using the same convenient mechanism for `bindings_set` and for exit reasons right now.

The simplest thing to do on the C side might be to get rid of the `exit_op_count`, and prefix all exit counters with `exit_`. So then, the toplevel exit counter for `send` would be `exit_send`. Then the exit reasons would be things like:

- `exit_send_callsite_not_simple`
- `exit_send_kw_splat`
- `exit_send_ic_empty`
- ...

The C code could probably handle, based on the counter name, separating the exit counters into different hashes in the struct returned by `get_stat_counters`. I think it would make sense to include things like the inlined/outlined machine code size in these stats/counters returned by `get_stat_counters` and have all the stats printing be done in Rubyland, as opposed to the current scheme where we do some printing in Ruby and some in C.

noahgibbs:
PR to convert vm_insns_count and rb_compiled_iseq_count to counters rather than freestanding variables: https://github.com/Shopify/yjit/pull/109

noahgibbs:
In the meeting you talked about wanting to convert some of the stuff done in C to Ruby. It looks like right now some of the end-of-process report is printed from Ruby and some from C. Specifically, in yjit_iface.c a method at_exit_print_stats is called on exit, which calls the Ruby method YJIT._print_stats(). It displays everything in the exit report from "***YJIT: Printing runtime counters from yjit.rb***" to "opt_aref exit reasons:" and the counter value(s).

[...]

PR #111, opened 2021-06-29
Add exits to yjit stats, convert at-exit reporting to Ruby

noahgibbs:
As discussed in https://github.com/Shopify/yjit/issues/107, this adds exit_op_count and inlined/outlined code to the returned YJIT stats. It also converts the existing exit stats report, with no changes, from C to Ruby.

I'm initially marking this as a draft because I'm seeing some small differences in the statistics themselves (e.g. number of side-exits from opt_send_without_block). Investigating now.

### DIFF
--- a/yjit_iface.c
+++ b/yjit_iface.c
@@ -766,10 +766,9 @@ get_yjit_stats(rb_execution_context_t *ec, VALUE self)
         // where the value is the count of side exits for that instruction.
 
         char key_string[rb_vm_max_insn_name_size + 6]; // Leave room for "exit_" and a final NUL
-        strncpy(key_string, "exit_", 6);               // Copy 6 characters to include the NUL.
         for (int i = 0; i < VM_INSTRUCTION_SIZE; i++) {
             const char *i_name = insn_name(i); // Look up Ruby's NUL-terminated insn name string
-            strncpy(key_string + 5, i_name, rb_vm_max_insn_name_size + 1);
+            snprintf(key_string, rb_vm_max_insn_name_size + 6, "%s%s", "exit_", i_name);
 
             VALUE key = ID2SYM(rb_intern(key_string));
             VALUE value = LL2NUM((long long)exit_op_count[i]);


yjit.rb
```diff
@@ -125,7 +125,7 @@ def self.disasm_block(cs, block, highlight)
   # Return nil when option is not passed or unavailable.
   def self.runtime_stats
     # defined in yjit_iface.c
-    Primitive.get_stat_counters
+    Primitive.get_yjit_stats
```
maximecb, line 128: Good call renaming this.

yjit_iface.c
```diff
@@ -747,6 +760,22 @@ get_stat_counters(rb_execution_context_t *ec, VALUE self)
             name_reader = name_end;
         }
     }
+
+    {
+        // Iterate through exit_op_count
```
maximecb, line 765: I think this needs a bit more commenting explaining what this is/does.

maximecb:
Yes that seems like an improvement 👍

noahgibbs:
I've changed the strcpy of a static string to strncpy to satisfy caution-tape-bot. I had fixed the potential variable-length strcpy earlier, but I think it still doesn't like me copying "exit_" with plain strcpy.

yjit_iface.c
```diff
@@ -747,6 +760,23 @@ get_stat_counters(rb_execution_context_t *ec, VALUE self)
...
+
+    {
+        // For each entry in exit_op_count, add a stats entry with key "exit_INSTRUCTION_NAME",
+        // where the value is the count of side exits for that instruction.
+
+        char key_string[rb_vm_max_insn_name_size + 6]; // Leave room for "exit_" and a final NUL
+        strncpy(key_string, "exit_", 6);               // Copy 6 characters to include the NUL.
+        for (int i = 0; i < VM_INSTRUCTION_SIZE; i++) {
+            const char *i_name = insn_name(i); // Look up Ruby's NUL-terminated insn name string
+            strncpy(key_string + 5, i_name, rb_vm_max_insn_name_size + 1);
```
jhawthorn, line 772: This looks right but might be simpler as `snprintf`, to avoid counting lengths by hand.

noahgibbs: That's a good idea. Done.

maximecb:
@noahgibbs is this PR ready to merge?

PR #112, opened 2021-06-30
Allow chaining on immediate guard

jhawthorn:
In `jit_guard_known_klass` whenever we encounter a new class should recompile the current instruction.

However, previously once `jit_guard_known_klass` had guarded for a heap object it would not recompile for  any immediate (special const) objects arriving afterwards and would take a plain side-exit instead of a chain guard.

[I made a gist demonstrating this issue](https://gist.github.com/jhawthorn/9d6280eae4ae21339ed63ba6ff624ece). When we see `nil` first and the heap object second we successfully generate 4 blocks and don't side exit. When we see the heap object first we never compile the version of the block for `nil` and side exit instead.

This is also demonstrated in a test in #103:
https://github.com/Shopify/yjit/blob/85be28e332fb4760461b3972862d870246913c89/test/ruby/test_yjit.rb#L58-L79

This commit uses `jit_chain_guard` inside `jit_guard_known_klass` instead of the plain side exit, so that we can recompile for any special constants arriving afterwards. With this change both versions of my test gist compile 4 blocks and have no side-exits.

---

I was a little unsure of this approach at first, having each of the three guards (`x & IMM_MASK`, `x <= Qnil`, `x.class == known`) jump to a different stub seemed wasteful. What I didn't grok initially (had a helpful hint from @maximecb) is that having the separate targets _does_ allow us to "learn" more information about the receiving value, even if we don't set explicit typing information in the branch (though we maybe could in the future? more later). Since we're only going to reach those stubs after going through the unique jump we implicitly get "information" from the values arriving to that stub and having their checks recompiled there (neat!).

(maybe rambling a bit here) As I mentioned, this doesn't set type information when branching, at this point I don't think anything would take advantage of it. In the future it feels like there's opportunity here, particularly with the comparison against `Qnil` could differentiate between heap values, Qnil, and Qfalse in one comparison and two jumps. We could also probably avoid the check after going through a `branchif` or `branchunless` if we recorded that that the type could not be a "falsy". This is all definitely out of scope for this PR though.

Another potential concern with this is that this is now a tree instead of a linear chain guarding the class, which might not fit the existing meaning of `SEND_MAX_DEPTH`? (or maybe it does? trees have depth) There's fortunately a small finite number of types of special constants. I think it's okay but maybe that needs to be made clearer as our behaviour.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -942,11 +942,29 @@ gen_jz_to_target0(codeblock_t *cb, uint8_t *target0, uint8_t *target1, uint8_t s
     }
 }
 
+static void
+gen_jbe_to_target0(codeblock_t *cb, uint8_t *target0, uint8_t *target1, uint8_t shape)
+{
+    switch (shape)
+    {
+        case SHAPE_NEXT0:
+        case SHAPE_NEXT1:
+        RUBY_ASSERT(false);
+        break;
+
+        case SHAPE_DEFAULT:
+        jbe_ptr(cb, target0);
+        break;
+    }
+}
+
 enum jcc_kinds {
     JCC_JNE,
     JCC_JNZ,
     JCC_JZ,
     JCC_JE,
+    JCC_JBE,
+    JCC_JNA,
 };
 
 // Generate a jump to a stub that recompiles the current YARV instruction on failure.
@@ -964,6 +982,10 @@ jit_chain_guard(enum jcc_kinds jcc, jitstate_t *jit, const ctx_t *ctx, uint8_t d
         case JCC_JZ:
         case JCC_JE:
             target0_gen_fn = gen_jz_to_target0;
+            break;
+        case JCC_JBE:
+        case JCC_JNA:
+            target0_gen_fn = gen_jbe_to_target0;
             break;
         default:
             RUBY_ASSERT(false && "unimplemented jump kind");
@@ -2258,9 +2280,9 @@ jit_guard_known_klass(jitstate_t *jit, ctx_t *ctx, VALUE known_klass, insn_opnd_
             ADD_COMMENT(cb, "guard not immediate");
             RUBY_ASSERT(Qfalse < Qnil);
             test(cb, REG0, imm_opnd(RUBY_IMMEDIATE_MASK));
-            jnz_ptr(cb, side_exit);
+            jit_chain_guard(JCC_JNZ, jit, ctx, max_chain_depth, side_exit);
             cmp(cb, REG0, imm_opnd(Qnil));
-            jbe_ptr(cb, side_exit);
+            jit_chain_guard(JCC_JBE, jit, ctx, max_chain_depth, side_exit);
 
             ctx_set_opnd_type(ctx, insn_opnd, TYPE_HEAP);
         }


maximecb:
> Another potential concern with this is that this is now a tree instead of a linear chain guarding the class, which might not fit the existing meaning of SEND_MAX_DEPTH

We did expect this to happen. I don't think it's generally a big concern because you only compile code for the types you actually see at run-time. The important part is just that there's some reasonable limit for the rare worst case scenarios.

jhawthorn:
> any idea how this affects our coverage numbers?

Looks like it moves "Ratio in YJIT" from 59.3% to 59.7%.


<details>
<summary>
<b>Before</b>
</summary>

```
$ ~/src/yjit/ruby --yjit-stats run_once.rb
**RUNNING BENCHMARK***
***YJIT: Printing runtime counters from yjit.rb***
Number of bindings allocated: 0
Number of locals modified through binding: 0
method call exit reasons:
      callsite_not_simple     170478 (43.2%)
         optimized_method      82429 (20.9%)
                  bmethod      76686 (19.4%)
      iseq_complex_callee      40059 (10.1%)
          ivar_set_method      12012 ( 3.0%)
       iseq_only_keywords      11217 ( 2.8%)
                 kw_splat       1978 ( 0.5%)
    cfunc_ruby_array_varg         16 ( 0.0%)
leave exit reasons:
      interp_return     951581 (64.0%)
    se_finish_frame     535002 (36.0%)
       se_interrupt          7 ( 0.0%)
getinstancevariable exit reasons:
              undef      12114 (69.6%)
    name_not_mapped       5282 (30.4%)
setinstancevariable exit reasons:
    (all relevant counters are zero)
opt_aref exit reasons:
    (all relevant counters are zero)
compiled_iseq_count:         1415
inline_code_size:         2058025
outlined_code_size:       1743193
total_exit_count:         2436345
total_insns_count:       80428463
vm_insns_count:          32763498
yjit_insns_count:        50101310
ratio_in_yjit:              59.3%
avg_len_in_yjit:             19.6
top-20 most frequent exit ops (95.4% of exits):
    opt_send_without_block:     549782 (22.6%)
                     leave:     535009 (22.0%)
               invokesuper:     408699 (16.8%)
                      send:     168784 (6.9%)
                splatarray:     166018 (6.8%)
                    opt_eq:      73045 (3.0%)
          getclassvariable:      71848 (2.9%)
               expandarray:      40766 (1.7%)
        getblockparamproxy:      40180 (1.6%)
               invokeblock:      36411 (1.5%)
        opt_getinlinecache:      33514 (1.4%)
                  opt_aset:      31948 (1.3%)
         opt_case_dispatch:      31694 (1.3%)
             getblockparam:      31226 (1.3%)
             setlocal_WC_1:      25457 (1.0%)
                 opt_nil_p:      21839 (0.9%)
                   opt_not:      17986 (0.7%)
       getinstancevariable:      14125 (0.6%)
                  opt_aref:      13992 (0.6%)
                   opt_mod:      12065 (0.5%)
```
</details>

<details>
<summary>
<b>After</b>
</summary>

```
$ ~/src/yjit/ruby --yjit-stats run_once.rb
***RUNNING BENCHMARK***
***YJIT: Printing runtime counters from yjit.rb***
Number of bindings allocated: 0
Number of locals modified through binding: 0
method call exit reasons:
      callsite_not_simple     170478 (43.2%)
         optimized_method      82429 (20.9%)
                  bmethod      76686 (19.4%)
      iseq_complex_callee      40059 (10.1%)
          ivar_set_method      12012 ( 3.0%)
       iseq_only_keywords      11217 ( 2.8%)
                 kw_splat       1978 ( 0.5%)
    cfunc_ruby_array_varg         16 ( 0.0%)
leave exit reasons:
      interp_return     945606 (63.9%)
    se_finish_frame     535134 (36.1%)
       se_interrupt          4 ( 0.0%)
getinstancevariable exit reasons:
              undef      12114 (69.6%)
    name_not_mapped       5282 (30.4%)
setinstancevariable exit reasons:
    (all relevant counters are zero)
opt_aref exit reasons:
    (all relevant counters are zero)
compiled_iseq_count:         1415
inline_code_size:         2068521
outlined_code_size:       1929353
total_exit_count:         2428343
total_insns_count:       80506158
vm_insns_count:          32479153
yjit_insns_count:        50455348
ratio_in_yjit:              59.7%
avg_len_in_yjit:             19.8
top-20 most frequent exit ops (95.6% of exits):
    opt_send_without_block:     541656 (22.3%)
                     leave:     535138 (22.0%)
               invokesuper:     412368 (17.0%)
                      send:     168784 (7.0%)
                splatarray:     166016 (6.8%)
                    opt_eq:      84945 (3.5%)
          getclassvariable:      71848 (3.0%)
               expandarray:      40768 (1.7%)
        getblockparamproxy:      40180 (1.7%)
               invokeblock:      36415 (1.5%)
        opt_getinlinecache:      33519 (1.4%)
                  opt_aset:      31948 (1.3%)
         opt_case_dispatch:      31694 (1.3%)
             getblockparam:      31226 (1.3%)
             setlocal_WC_1:      25457 (1.0%)
                 opt_nil_p:      18168 (0.7%)
       getinstancevariable:      14125 (0.6%)
                  opt_aref:      13992 (0.6%)
                   opt_mod:      12065 (0.5%)
                  opt_plus:      10593 (0.4%)
```
</details>

PR #113, opened 2021-06-30
Support jit_guard_known_klass on Symbols, Integer, and Floats

jhawthorn:
This adds support for guarding immediate (special const) Symbols, Integers, and Floats. These three classes can either be immediate or heap objects. This PR solves this by determining what kind we're seeing (based on the sample instance) and treating heap vs immediate essentially as though they were separate classes (requiring compiling a two versions if we happen to receive both kinds).

One thing I was unsure about is whether ex. `ETYPE_SYMBOL` is meant to refer to _only_ static symbols or if it could indicate a dynamic (on the heap) symbol as well. Because of this I ended up checking `is_imm` at compile time for each of these just in case.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1347,3 +1347,22 @@ assert_equal 'bar', %q{
   to_string((-"bar").to_sym)
   to_string((-"bar").to_sym)
 }
+
+# Call to flonum and heap float
+assert_equal '[nil, nil, nil, 1]', %q{
+  def is_inf(obj)
+    obj.infinite?
+  end
+
+  is_inf(0.0)
+  is_inf(0.0)
+  is_inf(1e256)
+  is_inf(1e256)
+
+  [
+    is_inf(0.0),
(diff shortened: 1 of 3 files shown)


maximecb:
We haven't implemented anything FP-related because it seems this most likely won't help the workloads we care about. However, symbols and integers guards are for sure worth implementing.

By static symbol the Ruby people mean immediate symbols? Any idea what proportion of symbols are static?

jhawthorn:
> We haven't implemented anything FP-related because it seems this most likely won't help the workloads we care about.

Agree, though I can think of cases it would be interesting to have, it's not something I would focus on. It was just the last case we weren't handling here. Let me know if you'd prefer that commit removed.

> By static symbol the Ruby people mean immediate symbols? Any idea what proportion of symbols are static?

Yeah! I made a [quick hack](https://github.com/Shopify/yjit/commit/39b9d420b7df4e248cbc3092f599cdce4ad82fae) and in here we're currently guarding against static symbols ~87k times and dynamic symbols ~4k times.

PR #125, opened 2021-07-13
Code page GC object

maximecb:


yjit_asm.h
```diff
@@ -242,8 +256,8 @@ x86opnd_t const_ptr_opnd(const void *ptr);
 
 // Machine code allocation
 uint8_t* alloc_exec_mem(uint32_t mem_size);
-uint8_t* alloc_code_page();
-void free_code_page(uint8_t* page_ptr);
+code_page_t* alloc_code_page();
```
XrXr, line 259: You probably want `(void)` here. Reasoning https://gist.github.com/XrXr/ad1dfe0cfb91dfc167672fef1c5f3157

yjit_iface.c
```diff
@@ -917,6 +923,35 @@ rb_yjit_iseq_free(const struct rb_iseq_constant_body *body)
...
+
+// Custom type for interacting with the GC
+static const rb_data_type_t yjit_code_page_type = {
+    "yjit_code_page",
+    {NULL, yjit_code_page_free, NULL, NULL},
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+// Allocate a code page and wrap it into a Ruby object owned by the GC
+VALUE rb_yjit_code_page_alloc()
```
XrXr, line 940: `(void)` here too. No codgen consequence here but good to keep consistent.

PR #127, opened 2021-07-15
Merge upstream Ruby commits

maximecb:


maximecb:
@jhawthorn wondering if you have any idea what might be causing a test to fail, if it's a problem with YJIT or something from upstream.

jhawthorn:
I don't have any idea. This is a strange one.

PR #128, opened 2021-07-15
Dup strings so we can mutate them

tenderlove:
I guess the strings are frozen, so we need to dup before mutating

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -103,7 +103,7 @@ module YJIT
     def self.disasm_block(cs, block, highlight)
       comments = comments_for(block.address, block.address + block.code.length)
       comment_idx = 0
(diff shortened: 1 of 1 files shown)


yjit.rb
```diff
@@ -63,7 +63,7 @@ def self.graphviz_for(iseq)
       highlight = ->(comment) { "<b>#{comment}</b>" }
       linebreak = "<br align=\"left\"/>\n"
 
-      buff = ''
+      buff = ''.dup
```
rafaelfranca, line 66:
```suggestion
      buff = +''
```

Since there is no pragma in the beginning of the file, the strings are only frozen if the command line option is passed, and the `+` method would only dup if they are frozen.

PR #130, opened 2021-07-15
Ensure we guard the value before we return

kddnewton:
Otherwise you can end up not implicitly calling `to_ary`, which if it has side-effects will result in different behavior.


XrXr (approved):
Thanks!

maximecb:
Nice catch.

PR #131, opened 2021-07-15
Use `jit_func` as YJIT's entry point

tenderlove:
This pull request changes the code to reuse MJIT's call handler as YJIT's entry point which should allow us to eliminate the scraper.

## Interpreter to YJIT calls

Rather than generating an instruction and having the call threaded interpreter call YJIT, we generate a function and set it as the `jit_func` on the iseq body.  When the interpreter calls `mjit_exec` YJIT gets the opportunity to compile an iseq and set the function pointer.  If the function pointer is already set, the interpreter will call it.

## YJIT to Interpreter Returns

Any time the interpreter calls in to MJIT generated function (or YJIT for that matter) it checks the return value of the generated function.  If the function returns `Qundef`, the interpreter will execute whatever CFP and PC is set on the EC.  If the function returns some other value, that other value will be used as the "return value" of the function and gets pushed on to the calling frame's stack.

MJIT uses `Qundef` to indicate "this function isn't done compiling yet", but we can easily use this value for side exits (which I've mentioned below).

### Successful YJIT returns

Successful YJIT calls will return some value besides `Qundef` to the interpreter.  However, this causes a little complication in the landing pad code.  Usually the `leave` instruction will push the return value on the top of the calling frame's stack.  But the interpreter is expecting the return value to be _returned_ from the JIT function, not manipulate the calling frame's stack.  YJIT's leave instruction doesn't know if it is returning to another iseq (a JIT to JIT return) or if it is returning to the interpreter.  So it always pushes a value on to the calling frame's stack.

The problem is that if YJIT returns to the interpreter and manipulates the calling frame's stack, we get a stack consistency error.  To handle this situation, the landing pad code pops the return value from the calling frame's SP, then returns the value.  This might be an area for improvement.

## YJIT to Interpreter Exits

If YJIT can't continue executing it can simply return `Qundef`.  Since the interpreter will begin executing the CFP at whatever PC is set, YJIT can adjust the PC and SP on the frame so that the interpreter will just pick up where YJIT left off.

## TODO

- [x] Make sure tests pass
- [x] Fix our register definitions. [They are backwards from the JIT entry point function](https://github.com/Shopify/yjit/pull/131/files#diff-746b53169139d506c1263176830553fb45c3a581c032e337f432bebb4754e2e7R367-R372)

~~Once this PR passes I'll send a follow up PR that removes the scraper.~~

I added a patch that removes the scraper.


maximecb:
Benchmark results for this branch on AWS:
```
-------------  -----------  ----------  ---------  ----------  -----------  ------------
bench          interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  yjit 1st itr
30k_ifelse     2366.4       0.0         392.2      0.2         6.03         4.61        
30k_methods    6423.1       0.0         870.3      0.1         7.38         6.98        
activerecord   163.7        0.5         131.1      0.6         1.25         1.20        
binarytrees    466.7        2.5         370.9      3.3         1.26         1.20        
cfunc_itself   103.4        0.3         57.1       1.0         1.81         1.78        
fannkuchredux  6546.6       0.1         6536.6     0.2         1.00         1.00        
fib            255.0        0.1         75.7       0.4         3.37         3.38        
getivar        109.0        0.1         44.3       0.6         2.46         1.00        
jekyll         9941.4       1.6         9275.2     1.9         1.07         1.07        
lee            1286.5       0.8         1134.9     1.5         1.13         1.12        
liquid-render  215.5        0.7         167.1      1.6         1.29         1.29        
nbody          126.5        0.1         124.7      0.3         1.01         1.01        
optcarrot      6391.7       0.9         3983.3     0.1         1.60         1.59        
psych-load     2551.5       0.1         2197.5     0.0         1.16         1.16        
railsbench     3805.9       0.9         3347.1     1.2         1.14         1.13        
respond_to     272.3        0.5         211.8      1.2         1.29         1.33        
setivar        86.3         0.1         56.3       0.5         1.53         1.00        
-------------  -----------  ----------  ---------  ----------  -----------  ------------
```

PR #134, opened 2021-07-20
Fix use-after-free on USE_EMBED_CI=0

XrXr:
The old code didn't keep old_operands[0] reachable while allocating. You
can crash it by requiring erb under GC stress mode.


XrXr:
This is also an issue upstream, we just happen to run into it because YJIT allocates more things and so triggers GC at different places.

noahgibbs:
Can we also submit the fix upstream?

maximecb:
Yeap yeap yeap. I should have updated JHawthorn's branch before merging it.

maximecb:
@XrXr unfortunately we're still getting some failures in the TypeProf gem.

XrXr:
The TypeProf issue also happens upstream. Maybe we want to merge in upstream once it's fixed there. https://github.com/ruby/ruby/pull/4663/checks?check_run_id=3115970208

PR #135, opened 2021-07-20
Fix and uncomment getlocal test

jhawthorn:
Previously, under the scraper, this would side-exit twice because it was returning to a C method (`each`). Now that we use the `jit_func` entrypoint, this test no longer side-exits 🎉.


maximecb:
It might not be getting executed tho:

```
total_exit_count:               0
total_insns_count:            413
vm_insns_count:               385
yjit_insns_count:              28
ratio_in_yjit:               6.8%
```

PR #136, opened 2021-07-20
Implement jitdump support

jhawthorn:
This adds support for outputting the [jitdump format for Linux perf](https://raw.githubusercontent.com/torvalds/linux/master/tools/perf/Documentation/jitdump-specification.txt). This allows perf to annotate and better display jitted code (with as few extra steps).

This works by writing an append-only file in the jitdump format, with a record for each block we compile. When we resize a block (ex. avoiding a trailing branch) we write it to the file again. This adds a monotonically increasing id to blocks to track them. We also must `mmap` the output file so that `perf inject` can later detect it (a bit odd, not well documented, figured this out from reading V8).

I just added the most basic support (the `JIT_CODE_LOAD` record), later we might want to add `JIT_CODE_DEBUG_INFO` records which could allow us to see the Ruby source inline in perf (which would be so cool!)

**Usage:**

Run `perf record` with CLOCK_MONOTONIC timer and tell yjit to emit the jitdump file.

```
perf record -k mono ./ruby --yjit-jitdump my_test_file.rb
```

Run perf inject to combine the jitdump and perf.data files (this creates a "fake" `.so` file for every block, so makes a bit of a mess)

```
perf inject -v -j -i perf.data -o perf.data.jitted
```

Run perf report on the new "injected" file.

```
perf report -i perf.data.jitted
```

**Before**

Screen Shot 2021-07-20 at 10 28 52 AM
Screen Shot 2021-07-20 at 10 58 35 AM

**After**

Screen Shot 2021-07-20 at 10 28 21 AM
Screen Shot 2021-07-20 at 10 57 34 AM

**References:**
* https://v8.dev/docs/linux-perf 
* https://github.com/v8/v8/blob/main/src/diagnostics/perf-jit.cc
* https://raw.githubusercontent.com/torvalds/linux/master/tools/perf/Documentation/jitdump-specification.txt

**TODO:**
* [x] Guards to only support this on linux
* [x] Fix build
* [x] Better buffering of output file?
* [x] Better names for methods?
* [ ] There are still some `[unknown]`s... maybe from the outlined code?

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -1062,6 +1062,7 @@ setup_yjit_options(const char *s, struct rb_yjit_options *yjit_opt)
         if (!yjit_opt->jitdump_file) {
             perror("fopen");
         }
+        setvbuf(yjit_opt->jitdump_file, NULL, _IOFBF, 2 * 1024 * 1024);
     }
 #endif
     else {


maximecb:
Looks interesting but it may be hard to interpret the results as is?

jhawthorn:
I think this is ready for review. Here's another example of measuring `fib.rb` from yjit-bench

**Before**

Screen Shot 2021-08-05 at 12 52 59 PM

**After**

Screen Shot 2021-08-05 at 12 52 46 PM

ruby.c
```diff
@@ -1054,6 +1054,17 @@ setup_yjit_options(const char *s, struct rb_yjit_options *yjit_opt)
     else if (opt_match_noarg(s, l, "test-backend")) {
         yjit_opt->test_backend = true;
     }
+#if YJIT_JITDUMP
```
maximecb, line 1057: May want to add that under the --yjit arguments?

yjit_core.c
```diff
@@ -621,11 +624,19 @@ void limit_block_versions(blockid_t blockid, ctx_t* ctx)
...
 }
 
+static block_t* allocate_block()
+{
+    static uint64_t last_unique_id = 0;
+    block_t* block = calloc(1, sizeof(block_t));
+    block->unique_id = ++last_unique_id;
+    return block;
+}
+
```
maximecb, line 634: How do static variables behave when there are multiple threads/ractors?

ruby.c
jhawthorn: I think this is, `--yjit-jitdump`. The macro is to test that it's available (ie. we are on linux)

PR #139, opened 2021-07-22
Preserve type mappings when manipulating the stack

jhawthorn:
Similar to the issues in #122 and #123, when manipulating the stack via `setn` and `swap` could end up "learning" the wrong types for the locals they came from.

To solve this, but to keep the mappings intact, this PR introduces `ctx_copy_opnd_mapping` and `ctx_swap_opnd_mapping` (I'm not sure this is the best interface, but I needed to be able to copy both mappings and sometimes type). I also use these for `dup` and `dupn` so that they maintain the mapping to the locals and we have more opportunity to learn types.

To help ensure types are tracked correctly, this also renames `ctx_set_opnd_type` to `ctx_upgrade_opnd_type` (also considered "ctx_learn_opnd_type", but "upgrade" sounded the clearest to me) to hopefully make the behaviour more obvious. `ctx_upgrade_opnd_type` now also includes assertions that the types being "upgraded" are actual upgrades (ie. we don't set an incompatible type or "unlearn" anything).

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -551,11 +551,11 @@ static codegen_status_t
 gen_dup(jitstate_t* jit, ctx_t* ctx)
 {
     // Get the top value and its type
-    val_type_t dup_type = ctx_get_opnd_type(ctx, OPND_STACK(0));
     x86opnd_t dup_val = ctx_stack_pop(ctx, 0);
+    temp_type_mapping_t mapping = ctx_get_opnd_mapping(ctx, OPND_STACK(0));
 
     // Push the same value on top
-    x86opnd_t loc0 = ctx_stack_push(ctx, dup_type);
+    x86opnd_t loc0 = ctx_stack_push_mapping(ctx, mapping);
     mov(cb, REG0, dup_val);
     mov(cb, loc0, REG0);
 
@@ -573,17 +573,16 @@ gen_dupn(jitstate_t* jit, ctx_t* ctx)
         return YJIT_CANT_COMPILE;
     }
 
-    val_type_t type1 = ctx_get_opnd_type(ctx, OPND_STACK(1));
     x86opnd_t opnd1 = ctx_stack_opnd(ctx, 1);
-
-    val_type_t type0 = ctx_get_opnd_type(ctx, OPND_STACK(0));
     x86opnd_t opnd0 = ctx_stack_opnd(ctx, 0);
+    temp_type_mapping_t mapping1 = ctx_get_opnd_mapping(ctx, OPND_STACK(1));
+    temp_type_mapping_t mapping0 = ctx_get_opnd_mapping(ctx, OPND_STACK(0));
 
-    x86opnd_t dst1 = ctx_stack_push(ctx, type1);
+    x86opnd_t dst1 = ctx_stack_push_mapping(ctx, mapping1);
     mov(cb, REG0, opnd1);
     mov(cb, dst1, REG0);
 
-    x86opnd_t dst0 = ctx_stack_push(ctx, type0);
+    x86opnd_t dst0 = ctx_stack_push_mapping(ctx, mapping0);
     mov(cb, REG0, opnd0);
     mov(cb, dst0, REG0);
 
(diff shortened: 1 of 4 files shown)


yjit_codegen.c
```diff
@@ -1664,6 +1659,26 @@ guard_two_fixnums(ctx_t* ctx, uint8_t* side_exit)
...
+
+    if (arg0_type.type != ETYPE_FIXNUM && arg0_type.type != ETYPE_UNKNOWN) {
+        jmp_ptr(cb, side_exit);
+        return;
+    }
+
+    if (arg1_type.type != ETYPE_FIXNUM && arg1_type.type != ETYPE_UNKNOWN) {
+        jmp_ptr(cb, side_exit);
+        return;
+    }
```
jhawthorn, line 1675: IMO it should be invalid to call `guard_two_fixnums` in this situation, but this changeset was already getting a bit long, so I'd like to address that in a separate PR.

yjit_core.c
```diff
@@ -172,20 +179,63 @@ void ctx_set_opnd_type(ctx_t* ctx, insn_opnd_t opnd, val_type_t type)
...
+
+    ctx->temp_types[idx0] = ctx->temp_types[idx1];
+    ctx->temp_mapping[idx0] = ctx->temp_mapping[idx1];
+
+    ctx->temp_types[idx1] = temp_type;
+    ctx->temp_mapping[idx1] = temp_mapping;
+}
+
+void
+ctx_copy_opnd_mapping(ctx_t* ctx, insn_opnd_t dest, insn_opnd_t src)
```
maximecb, line 223: I would like to see comments explaining what these methods do and why they exist, to help current and future team members understand.

maximecb:
Are you purely trying to improve how much info we retain here or were we not not handling something correctly?

yjit_core.c
```diff
@@ -153,13 +153,20 @@ ctx_get_opnd_type(const ctx_t* ctx, insn_opnd_t opnd)
...
+    RUBY_ASSERT(type_diff((src), (dest)) != INT_MAX); \
+    (dest) = (src); \
+} while (false)
+
+
 /**
 Set the type of an instruction operand
 */
-void ctx_set_opnd_type(ctx_t* ctx, insn_opnd_t opnd, val_type_t type)
+void ctx_upgrade_opnd_type(ctx_t* ctx, insn_opnd_t opnd, val_type_t type)
```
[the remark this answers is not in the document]
maximecb: This is a good idea. I appreciate you taking the time to make improvements like this to the type info tracking.

jhawthorn:
> Can you expand on what was being handled incorrectly before?

Here's an example I found though taking the assertions added in this PR, but not the changes to swap.

`self.call_count += 1` ([rubyexplorer.xyz YARV bytecode](
https://www.rubyexplorer.xyz/explores?code=self.call_count+%2B%3D+1&coverage_enabled=true&debug_frozen_string_literal=false&frozen_string_literal=false&inline_const_cache=true&instructions_unification=false&operands_unification=true&peephole_optimization=true&specialized_instruction=true&stack_caching=false&tailcall_optimization=false))

```
0000 putself
0001 dup
0002 opt_send_without_block <calldata!mid:call_count, argc:0, FCALL|ARGS_SIMPLE>
0004 putobject_INT2FIX_1_
0005 opt_plus <calldata!mid:+, argc:1, ARGS_SIMPLE>
0007 swap
0008 topn 1
0010 opt_send_without_block <calldata!mid:call_count=, argc:1, FCALL|ARGS_SIMPLE>
0012 pop
0013 leave
```

At position 7, right before we call swap, our stack will be in the state
| Mapping | Type | |
|--------|--------|---|
| TEMP_SELF | - | From `putself` instruction |
| TEMP_STACK | FIXNUM | Result of adding `1` to the value returned by `call_count` |

(Prior to this PR we would not know the type of `self` here since our call would have guarded on the temporary dup-ed value, this issue with swap exists either way)

Previously, without changing the mappings, `swap` would call `ctx_set_opnd_type` on each of these. This is fine for the `TEMP_STACK` mapping, but for the `TEMP_SELF` mapping we would incorrectly "learn" that self is a FIXNUM.

maximecb:
That makes sense. Good catch.

yjit_core.c
```diff
@@ -154,39 +158,92 @@ ctx_get_opnd_type(const ctx_t* ctx, insn_opnd_t opnd)
...
         case TEMP_LOCAL:
         RUBY_ASSERT(mapping.idx < MAX_LOCAL_TYPES);
-        ctx->local_types[mapping.idx] = type;
+        UPGRADE_TYPE(ctx->local_types[mapping.idx], type);
         break;
     }
 }
 
+temp_type_mapping_t
+ctx_get_opnd_mapping(const ctx_t* ctx, insn_opnd_t opnd)
```
maximecb, line 203: I'd like to see some descriptive comments above every new method explaining the what they do and why we need them, to make the codebase more friendly for new people.

jhawthorn:
Okay! I think this is ready for review. I found a few more bugs and wrote a regression test.

I ended up replacing the previous, awkward API (copy and swap) with one more similar to the existing get/set. This ended up allowing adding a `ctx_push_opnd_mapping` and refactoring all the existing types of pushes into one 🎉.

yjit_core.c
```diff
@@ -154,39 +156,104 @@ ctx_get_opnd_type(const ctx_t* ctx, insn_opnd_t opnd)
...
 {
     if (opnd.is_self) {
-        ctx->self_type = type;
+        UPGRADE_TYPE(ctx->self_type, type);
         return;
     }
 
-    if (ctx->stack_size > MAX_TEMP_TYPES)
+    if (ctx->stack_size >= MAX_TEMP_TYPES)
         return;
```
XrXr, line 180: Not related to this PR, but I'm not sure why we need to return in this case. We can still upgrade when the stack is large no?

XrXr: Also, I think the old condition was fine because of the -1 in the index calculation below?

jhawthorn: 👍 I must have misread this as checking the index into `temp_mapping`, rather than the total stack size. Pushed a commit to check `stack_idx` instead (so allowing upgrades of first N types).

maximecb:
Good work John!

Issue #140, opened 2021-07-27
Segfault in YJIT.runtime_stats with --disable-yjit


noahgibbs:
This is occurring without YJIT active (disabled by command line flag), in a YJIT codebase compiled without RUBY_DEBUG.

I assume cb and/or ocb aren't being set up when YJIT is disabled. So I should check for that and not return a stats hash (or return an empty stats hash) when YJIT is disabled.

noahgibbs:
They're equally easy to check for. You either know it's a hash and check with empty?, or you know it's hash-or-nil and check with nil? or ==nil. I'm not sure if either way is more semantically sensible, and I'm pretty sure neither is noticeably easier.

So, y'know, no opinion between them.

maximecb:
Let's go for nil since we're not collecting any stats and it's kind of an invalid operation to request them when YJIT is disabled.

PR #141, opened 2021-07-27
Don't add code size YJIT-stats if codeblock pointer is NULL.

noahgibbs:
Fix issue #140


maximecb (approved):

noahgibbs:
I'll rebase onto main and see if I keep seeing this CI error.

PR #142, opened 2021-07-27
Add getglobal to yjit

eileencodes:
Adds getglobal to yjit and a test for it.

Co-authoredby: Aaron Patterson <tenderlove@ruby-lang.org>


yjit_codegen.c
```diff
@@ -3423,6 +3424,27 @@ gen_leave(jitstate_t* jit, ctx_t* ctx)
...
+gen_getglobal(jitstate_t* jit, ctx_t* ctx)
+{
+    ID gid = jit_get_arg(jit, 0);
+
+    // Save YJIT registers
+    yjit_save_regs(cb);
+
+    mov(cb, RDI, imm_opnd(gid));
+
+    call_ptr(cb, RSI, (void *)&rb_gvar_get);
```
maximecb, line 3437: Looks pretty good but I would use the `C_ARG_REGS` definition for the `gid` instead of directly using `RDI`.

For call_ptr you should use `REG0` as a temporary register instead of directly using `RSI`.

eileencodes: Ok this is my first assembly code so I think I got it right but let me know if I didn't 😄

maximecb: No problem :)

PR #144, opened 2021-07-28
Add setglobal to yjit

eileencodes:
Adds yjit support for setting global variables.

Co-authored-by: Aaron Patterson <tenderlove@ruby-lang.org>
Co-authored-by: John Hawthorn <john@hawthorn.email>

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -3446,6 +3446,28 @@ gen_getglobal(jitstate_t* jit, ctx_t* ctx)
 }
 
 static codegen_status_t
+gen_setglobal(jitstate_t* jit, ctx_t* ctx)
+{
+    ID gid = jit_get_arg(jit, 0);
+
+    // Save YJIT registers
+    yjit_save_regs(cb);
+
+    mov(cb, C_ARG_REGS[0], imm_opnd(gid));
+
+    x86opnd_t val = ctx_stack_pop(ctx, 1);
+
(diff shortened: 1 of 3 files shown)


yjit_codegen.c
```diff
@@ -3445,6 +3445,28 @@ gen_getglobal(jitstate_t* jit, ctx_t* ctx)
...
+    ID gid = jit_get_arg(jit, 0);
+
+    // Save YJIT registers
+    yjit_save_regs(cb);
+
+    mov(cb, C_ARG_REGS[0], imm_opnd(gid));
+
+    x86opnd_t val = ctx_stack_pop(ctx, 1);
+
+    mov(cb, REG0, val);
```
maximecb, line 3460: I think you probably want to move `val` into `C_ARG_REGS[1]`

yjit_codegen.c
```diff
@@ -3445,6 +3445,28 @@ gen_getglobal(jitstate_t* jit, ctx_t* ctx)
...
+
+    call_ptr(cb, RAX, (void *)&rb_gvar_set);
```
maximecb, line 3462: And here the RAX should be REG0 (but it doesn't affect how the program works)

yjit_codegen.c
eileencodes: Whoops, fixed that and the one below.

PR #145, opened 2021-07-29
Merge upstream Ruby commits

maximecb:
Will it merge...?


maximecb:
I wonder if tests are failing here because of the bugs you found in the type info tracking @jhawthorn 🤔

jhawthorn:
Yes. That's 100% a bug I saw there in URI::Generic. Specifically this should fix it https://github.com/Shopify/yjit/pull/139/commits/c8c4e67c6b11da98f1a9a53a4c750e4b712de21b which I have a regression test for locally. I've been working on an upstream Ruby bug today, but I'll try to finish that PR up this afternoon 😁.

PR #146, opened 2021-07-29
Get some tests passing with the new backend

kddnewton:


yjit_backend.c
```diff
@@ -39,11 +39,13 @@ ir_opnd_t ir_mem(uint8_t num_bits, ir_opnd_t base, int32_t disp)
...
 }
 
+typedef rb_darray(int32_t) live_ranges_t;
+
 // Fake/temporaty JIT state object for testing/experimenting
 typedef struct yjit_jit_state
 {
     insn_array_t insns;
-
+    live_ranges_t live_ranges;
```
maximecb, line 48: We have to be careful with dynamic allocations in C. If you allocate this here then we have to free it somewhere. There's also some overhead in allocating this every time we compile a block. For now, I do think we probably want to prioritize convenience over speed while we develop the new backend, but it's something to keep in mind.

maximecb (approved):

PR #148, opened 2021-07-29
Get register allocation working for the new backend

kddnewton:


yjit_backend.c
```diff
@@ -217,13 +217,14 @@ void ir_print_insns(jitstate_t *jit)
 /* Convenience methods for pushing instructions. */
 /*************************************************/
 
-ir_opnd_t ir_add(jitstate_t *jit, ir_opnd_t dest, ir_opnd_t src)
+ir_opnd_t ir_add(jitstate_t *jit, ir_opnd_t left, ir_opnd_t right)
 {
-    return push_insn(jit, OP_ADD, dest, src);
+    return push_insn(jit, OP_ADD, left, right);
```
maximecb, line 222: I'd call them opnd0, opnd1, ...

yjit_backend.c
```diff
@@ -4,8 +4,8 @@
 #include "yjit_backend.h"
 
 // List of scratch registers
-#define NUM_SCR_REGS 2
-#define SCR_REGS (ir_opnd_t[]){ IR_REG(RAX), IR_REG(RCX) }
+#define NUM_SCR_REGS 3
```
maximecb, line 7: Can we get away with just 2 scratch registers? That leaves more for allocation?

maximecb:
Looks pretty good.

MSVC has decided that it doesn't like something about the C code. Happens sometimes 😅 

My main feedback would be: this probably needs a few more tests, and in particular, can you think of edge cases?

yjit_backend.c
kddnewton: 👍

yjit_backend.c
```diff
@@ -232,61 +233,120 @@ void ir_ret(jitstate_t *jit)
...
+        if (active[index] == NOT_LIVE_REG) {
             return index;
         }
     }
     RUBY_ASSERT(false && "out of free registers");
 }
 
 // Write out x86 instructions into the codeblock for the given IR instructions
-void ir_x86_insns(codeblock_t *cb, insn_array_t insns)
+void ir_x86_insns(codeblock_t *cb, jitstate_t *jit)
```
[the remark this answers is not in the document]
kddnewton: 👍

yjit_backend.c
kddnewton: Yup! I had to revisit the way I was handling `ir_mov` but it works with just 2 now.

yjit_backend.c
```diff
@@ -213,109 +213,203 @@ void ir_print_insns(jitstate_t *jit)
...
+int32_t ir_next_scratch(int32_t active[NUM_SCR_REGS])
 {
     for (int32_t index = 0; index < NUM_SCR_REGS; index++) {
-        if (!active[index]) {
+        if (active[index] == NOT_LIVE_REG) {
             return index;
         }
     }
     RUBY_ASSERT(false && "out of free registers");
 }
```
maximecb, line 293: Where does the register marked as active in this function?

yjit_backend.c
```diff
@@ -367,6 +461,17 @@ void test_backend()
...
+        ir_opnd_t opnd1 = ir_mov(jit, IR_REG(RAX), ir_imm(3));
+        ir_mov(jit, IR_REG(RAX), ir_add(jit, ir_imm(4), opnd1));
+        ir_ret(jit);
+    })
+
+    TEST("adding with both immediates", 7, {
+        ir_mov(jit, IR_REG(RAX), ir_add(jit, ir_imm(3), ir_imm(4)));
+        ir_ret(jit);
+    })
+
```
maximecb, line 474: Could you add a third test with two operations here? Are you able to handle that at this point?

Eg: RAX = 3 + 7 - 1

yjit_backend.c
```diff
@@ -213,109 +213,203 @@ void ir_print_insns(jitstate_t *jit)
...
+    printf("}\n");
+}
+
 /*************************************************/
 /* Convenience methods for pushing instructions. */
 /*************************************************/
 
-ir_opnd_t ir_add(jitstate_t *jit, ir_opnd_t dest, ir_opnd_t src)
+ir_opnd_t ir_add(jitstate_t *jit, ir_opnd_t opnd1, ir_opnd_t opnd2)
```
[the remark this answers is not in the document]
kddnewton: 👍

yjit_backend.c
maximecb: Just asking because the comment says "mark it as active". It should probably say "find the next free scratch register" instead.

maximecb (approved):
Good job Kevin! :)

PR #149, opened 2021-08-02
Implement ir_and, ir_not, ir_sub

kddnewton:
Works by moving the `add` logic into a function that accepts a pointer to an x86 writer function, then calls the same kind of code for each operation type.

### DIFF
--- a/yjit_backend.c
+++ b/yjit_backend.c
@@ -171,8 +171,11 @@ const char* ir_op_name(int op)
     switch (op)
     {
         case OP_ADD: return "add";
+        case OP_AND: return "and";
         case OP_MOV: return "mov";
+        case OP_NOT: return "not";
         case OP_RET: return "ret";
+        case OP_SUB: return "sub";
 
         default:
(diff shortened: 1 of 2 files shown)


yjit_backend.c
```diff
@@ -303,6 +316,47 @@ x86opnd_t ir_alloc_reg(int32_t *active, x86opnd_t *allocations, int32_t insn_idx
...
+    x86opnd_t opnd0 = ir_x86opnd(allocations, rb_darray_get(insn.opnds, 0));
+    x86opnd_t opnd1 = ir_x86opnd(allocations, rb_darray_get(insn.opnds, 1));
+
+    if (opnd0.type == OPND_REG) {
+        // If the first operand is already a register, then use that as the
+        // destination for the instruction.
+        op(cb, opnd0, opnd1);
+
+        // If this instruction persists through other instructions, then make
+        // sure we move the result into the correct register.
```
maximecb, line 333: Not sure what you mean here by persists through other instructions?

kddnewton:
That makes sense! I'll give it another go

PR #151, opened 2021-08-03
Add verify_ctx to validate learned types against received types

jhawthorn:
An earlier version of this helped track down some of the issues fixed by https://github.com/Shopify/yjit/pull/139

This adds a `verify_ctx` method, which, when RUBY_DEBUG is enabled, checks that the detected types of the values on the stack, locals, and self are compatible with the actual compile-time types. I think this could still provide some value in the future to ensure our type bookkeeping is correct (at least against the values we see at compile-time).

This also found an issue (though I'm not sure the code could ever be executed due to the pc_is_zero guard) where when encountering optargs we could call `yjit_gen_block` with a `starting_insn_idx` which did not match our actual current PC from ec.

For this to work `jit_type_of_value` needs to find as specific types as the other guards do (which it should be able to, as it has the actual current value) so that has been updated (which we probably want anyways 🙂)

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -127,9 +127,17 @@ jit_type_of_value(VALUE val)
             return TYPE_FIXNUM;
         } else if (NIL_P(val)) {
             return TYPE_NIL;
+        } else if (val == Qtrue) {
+            return TYPE_TRUE;
+        } else if (val == Qfalse) {
+            return TYPE_FALSE;
+        } else if (STATIC_SYM_P(val)) {
+            return TYPE_STATIC_SYMBOL;
+        } else if (FLONUM_P(val)) {
+            return TYPE_FLONUM;
         } else {
-            // generic immediate
-            return TYPE_IMM;
+            RUBY_ASSERT(false);
+            UNREACHABLE_RETURN(TYPE_IMM);
         }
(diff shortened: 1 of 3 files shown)


yjit_codegen.c
```diff
@@ -118,6 +118,18 @@ jit_peek_at_self(jitstate_t *jit, ctx_t *ctx)
...
+    int32_t local_table_size = jit->iseq->body->local_table_size;
+    RUBY_ASSERT(n < (int)jit->iseq->body->local_table_size);
+
+    const VALUE *ep = jit->ec->cfp->ep;
+    return ep[-VM_ENV_DATA_SIZE - local_table_size + n + 1];
+}
+
 // When we know a VALUE to be static, this returns an appropriate val_type_t
 static val_type_t
 jit_type_of_value(VALUE val)
```
maximecb, line 135: I think this function would probably belong in `yjit_core.c`, but it seems like a very useful addition.

maximecb:
This is a good idea. More testing is key to our success :)

yjit_codegen.c
jhawthorn: 👍 Was beginning to think that. Thanks for the confirmation

PR #158, opened 2021-08-11
Separate register allocation from code generation

kddnewton:
in the new backend.


maximecb:
Seems like a good start 👍

PR #165, opened 2021-08-12
Add toregexp to yjit

eileencodes:
The FIXME is there so we remember to investigate why insns clears the
temporary array. Is this necessary? If it's not we can remove it from
both.

Co-authored-by: Aaron Patterson <tenderlove@ruby-lang.org>


yjit_codegen.c
```diff
@@ -3511,6 +3512,47 @@ gen_tostring(jitstate_t* jit, ctx_t* ctx)
...
+
+    mov(cb, C_ARG_REGS[0], imm_opnd(0));
+    mov(cb, C_ARG_REGS[1], imm_opnd(cnt));
+    lea(cb, C_ARG_REGS[2], values_ptr);
+    call_ptr(cb, REG0, (void *)&rb_ary_tmp_new_from_values);
+
+    // Save the array so we can clear it later
+    push(cb, RAX);
+    push(cb, RAX); // Alignment
+    mov(cb, C_ARG_REGS[0], RAX);
```
maximecb, line 3537: RAX should probably be REG0 here

yjit_codegen.c
```diff
@@ -3511,6 +3512,47 @@ gen_tostring(jitstate_t* jit, ctx_t* ctx)
...
 }
 
+static codegen_status_t
+gen_toregexp(jitstate_t* jit, ctx_t* ctx)
+{
+    rb_num_t opt = jit_get_arg(jit, 0);
+    rb_num_t cnt = jit_get_arg(jit, 1);
+
+    // Save the PC and SP because we might make a Ruby call for
+    // Kernel#set_trace_var
```
maximecb, line 3522: set_trace_var ?

yjit_codegen.c
jhawthorn: I think `RAX` could be appropriate here since we are passing the return of the previous C call.

yjit_codegen.c
```diff
@@ -3511,6 +3512,47 @@ gen_tostring(jitstate_t* jit, ctx_t* ctx)
...
+    // rb_ary_tmp_new_from_values into C arg regs so we can clear it
+    pop(cb, REG1); // Alignment
+    pop(cb, C_ARG_REGS[0]);
+
+    // The value we want to push on the stack is in RAX right now
+    x86opnd_t stack_ret = ctx_stack_push(ctx, TYPE_UNKNOWN);
+    mov(cb, stack_ret, RAX);
+
+    // Clear the temp array. FIXME: is this necessary?
+    call_ptr(cb, REG0, (void *)&rb_ary_clear);
```
maximecb, line 3551: Is it necessary? 😅

eileencodes: Yea - we don't know if this is necessary in the original (I noted that in the commit message)

PR #174, opened 2021-08-23
Add advance_nocrashy_tag script

noahgibbs:
If this all works correctly, "nocrashy" should point to the most recent top-level merge commit where three merge commits in a row have all fully passed all GHActions. The tag should be updated twice daily.

### DIFF
--- a/.github/workflows/advance_nocrashy.yml
+++ b/.github/workflows/advance_nocrashy.yml
@@ -1,0 +1,14 @@
+name: Advance the "nocrashy" tag
+on:
+  schedule:
+    # Run twice daily at 5:47 and 11:47
+    - cron: '47 5,11 * * *'
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Advance the tag
+        run: |
+          ruby misc/advance_nocrashy_tag.rb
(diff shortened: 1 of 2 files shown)


noahgibbs:
Huh. I had wondered how GHActions in non-main branches worked -- its seems naively, for instance, that if you had 100 branches and a cron action, you wouldn't want to run 100 copies of it at a specific time of day.

GitHub appears to handle that by... not running GHActions that are in non-main branches. If you go to "actions" in the repo, for instance, you won't see the one that I added to this branch. That mostly makes sense as a solution, but it's likely to make it hard to debug this before it's merged, which is mixed :-/

maximecb:
> If this all works correctly, "nocrashy" should point to the most recent top-level merge commit where three merge commits in a row have all fully passed all GHActions. The tag should be updated twice daily.

I would prefer the least-recent commit where 3 merge commits have passed, because the later tests add extra validation that we didn't have a heisenbug before.

> GitHub appears to handle that by... not running GHActions that are in non-main branches. If you go to "actions" in the repo, for instance, you won't see the one that I added to this branch. That mostly makes sense as a solution, but it's likely to make it hard to debug this before it's merged, which is mixed :-/

Hmm that's weird. I would have thought that if you did a PR on main it would run your new actions. Seemed to be that way when I added the asm tests action. If you have to merge this on master to test it though, that's ok, we can always revert 👍

noahgibbs:
Hm. This is a draft PR, so maybe that's the problem. I'll mark it non-draft.

maximecb (approved):

PR #177, opened 2021-08-24
Allow to compile with --yjit-stats support but not the full RUBY_DEBUG

casperisfine:
RUBY_DEBUG have a very significant performance overhead. Enough that YJIT with RUBY_DEBUG is noticeably slower than the interpreter without RUBY_DEBUG.

This makes it hard to collect yjit-stats in production environments.

By allowing to collect JIT statistics without the RUBy_DEBUG overhead, I hope to make such use cases smoother.

WARNING: This is me playing :science-dog:, this PR is likely all wrong.

### DIFF
--- a/yjit_backend.c
+++ b/yjit_backend.c
@@ -25,6 +25,15 @@ ir_opnd_t ir_const_ptr(void *ptr)
     };
 }
 
+ir_opnd_t ir_str_ptr(char *str)
+{
+    return (ir_opnd_t) {
+        .num_bits = sig_imm_size((uint64_t) str),
+        .kind = EIR_IMM,
+        .as.str = str
+    };
+}
+
 // Immediate operand
 ir_opnd_t ir_imm(int64_t val)
 {
@@ -48,9 +57,18 @@ ir_opnd_t ir_mem(uint8_t num_bits, ir_opnd_t base, int32_t disp)
     };
 }
 
+ir_opnd_t ir_label_opnd(char *name)
+{
+    return (ir_opnd_t) {
+        .kind = EIR_LABEL_NAME,
+        .as.str = name
+    };
+}
+
 typedef rb_darray(int32_t) live_ranges_t;
 
 // Fake/temporary JIT state object for testing/experimenting
+// This object only exists during code generation
 typedef struct yjit_jit_state
 {
     // The current list of instructions that will be used to write out the
@@ -66,7 +84,7 @@ typedef struct yjit_jit_state
     live_ranges_t live_ranges;
 } jitstate_t;
 
-ir_opnd_t push_insn_variadic(jitstate_t* jit, ...)
+ir_opnd_t ir_push_insn_variadic(jitstate_t* jit, ...)
 {
     // Start VA parsing after last named argument
     va_list args;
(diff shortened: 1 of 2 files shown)


yjit_codegen.c
```diff
@@ -157,30 +157,7 @@ jit_save_sp(jitstate_t* jit, ctx_t* ctx)
...
-_counted_side_exit(uint8_t *existing_side_exit, int64_t *counter)
-{
-    if (!rb_yjit_opts.gen_stats) return existing_side_exit;
-
-    uint8_t *start = cb_get_ptr(ocb, ocb->write_pos);
-    _gen_counter_inc(ocb, counter);
-    jmp_ptr(ocb, existing_side_exit);
-    return start;
-}
+# define YJIT_STATS 1
```
noahgibbs, line 160: Not sure if we want an #ifndef around this. I can't think of many cases where we'd want to turn off YJIT_STATS on a RUBY_DEBUG build, but I feel kind of funny assuming we never would.

noahgibbs:
I'm seeing a line in yjit_core.c where the counters are directly incremented ("yjit_runtime_counters.compiled_iseq_count++;") that it looks like you missed - so it'd only be incremented on RUBY_DEBUG, not YJIT_STATS, which isn't what we want.

noahgibbs:
And it looks like you left it as RUBY_DEBUG in reset_stats_bang - that should be YJIT_STATS too.

noahgibbs:
Okay, guess I just looked right past reset_stats_bang in yours. Thought I looked for it and didn't find it, but I was probably just careless :-) So then that's fine!

yjit.rb
```diff
@@ -147,8 +147,8 @@ def _print_stats
       return unless stats
 
       $stderr.puts("***YJIT: Printing YJIT statistics on exit***")
-      $stderr.puts("Number of bindings allocated: %d\n" % stats[:binding_allocations])
-      $stderr.puts("Number of locals modified through binding: %d\n" % stats[:binding_set])
+      $stderr.puts("Number of bindings allocated: %d\n" % stats[:binding_allocations].to_i)
+      $stderr.puts("Number of locals modified through binding: %d\n" % stats[:binding_set].to_i)
```
maximecb, line 151: Curious why you added all the `.to_i` calls in this file? Are these values not already integers?

yjit_codegen.c
```diff
@@ -300,7 +313,7 @@ yjit_gen_exit(jitstate_t *jit, ctx_t *ctx, codeblock_t *cb)
     mov(cb, member_opnd(REG_CFP, rb_control_frame_t, pc), RAX);
 
     // Accumulate stats about interpreter exits
-#if RUBY_DEBUG
+#if RUBY_DEBUG || YJIT_STATS
```
maximecb, line 316: This should probably be #if YJIT_STATS, no reason to make it an or, unless I'm missing something.

casperisfine: Good catch, I fixed it.

yjit.rb
casperisfine: Right, I reverted it. It was during development, when I only had a partial solution, some of theses keys were missing and the code would raise. The `to_i` helped me figure out what was going on, but it wasn't meant to stick in the final PR.

maximecb (approved):
Nice work!

casperisfine:
Not sure what's going on, look like CI is broken:

```
Run mkdir build
mkdir: cannot create directory 'build': Permission denied
```

maximecb:
Not sure either 😨

casperisfine:
Looks like a GitHub Actions change: https://github.com/ruby/ruby/commit/2be84afea5167343e66be85b4cd7945533e4a7b5#diff-3e56f5a621f6c8bf5cac7ff51f771ab470d2e71216fe77c417b1622f782567f7

casperisfine:
This should do: https://github.com/Shopify/yjit/pull/179

PR #191, opened 2021-08-27
Add a Ruby wrapper object for each block_t allocated

tenderlove:
This commit adds a Ruby wrapper object for each `block_t` that is
allocated.  The wrapper object points at the `block_t`, and the
`block_t` points at it's own wrapper.

The mark function for the wrapper object will iterate through its
children (outgoing blocks) and mark the wrapper objects for those
blocks.

A new function `rb_yjit_mark_iseq_entry_blocks` is introduced that marks
only the entry blocks associated with the iseq passed in.  The entry
blocks will naturally mark their children, which will mark their
children, etc.

`block_t` wrapper objects also mark a "code page" reference, but
currently that reference is always nil.  Once we introduce code page
objects, the block wrapper objects will keep the code page objects
alive.

I gave up on doing this without `block_t` wrapper objects.  I still think it's possible, but I wanted to get something that works, then optimize later.

### DIFF
--- a/yjit_core.c
+++ b/yjit_core.c
@@ -640,7 +640,12 @@ block_t* gen_block_version(blockid_t blockid, const ctx_t* start_ctx, rb_executi
     // Allocate a new block version object
     block_t* block = calloc(1, sizeof(block_t));
     block->blockid = blockid;
+    block->self = yjit_wrap_block(block);
+
+    RB_OBJ_WRITTEN(block->blockid.iseq, Qundef, block->self);
+    block->code_page = Qnil;
     memcpy(&block->ctx, start_ctx, sizeof(ctx_t));
+
 
     // Store a pointer to the first block (returned by this function)
     block_t* first_block = block;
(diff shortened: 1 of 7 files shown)


yjit_iface.c
```diff
@@ -500,6 +534,67 @@ struct yjit_block_itr {
...
+
+        rb_darray_for(versions, block_idx) {
+            block_t *block = rb_darray_get(versions, block_idx);
+
+            if (block->blockid.idx == 0) {
+                cb(block, ctx);
+            }
+        }
+    }
+}
```
maximecb, line 553: AFAIK the array is indexed by position in the iseq, so you wouldn't need to iterate through every possible block_idx.

tenderlove: Ah ya, 🤦🏻‍♂️.  I'll fix it to just use the index and that should clean this function up.  Thank you!

yjit_core.c
```diff
@@ -640,6 +640,8 @@ block_t* gen_block_version(blockid_t blockid, const ctx_t* start_ctx, rb_executi
     // Allocate a new block version object
     block_t* block = calloc(1, sizeof(block_t));
     block->blockid = blockid;
+    block->self = yjit_wrap_block(block);
```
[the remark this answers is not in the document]
tenderlove: Ya, I agree.  I've made the `block_t` and Ruby wrapper objects point at each other.  I think we could remove the `block_t` -> Ruby pointer (the `self` pointer), but we probably have to store the Ruby object in the darray, then have reads from the darray unwrap the Ruby object.  Not impossible, just work.

PR #193, opened 2021-08-30
Shave a few instructions off of leave

XrXr:
The code path for leave that returns to the interpreter
(gen_leave() -> yjit_gen_leave_exit()) used to have the logic:

```
    cfp->sp++;
    cfp->sp[-1] = return_val;
    cfp->sp--;
    return return_val;
```

The SP changes it made was unnecessary and this change removes it.

After this change, `leave` doesn't adjust the `cfp->sp` of the caller
and only writes `cfp->sp[0]`. To accomodate this in the JIT-to-JIT
return case, return stubs have an `sp_offset` of 1.

The change removes sp adjustment from the JIT-to-JIT return case, too,
making it more efficient. Also, since the C method case of `send`
has an `sp_offset` of 1 after the call, this change enables block
version sharing.

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -912,6 +912,28 @@ gen_splatarray(jitstate_t* jit, ctx_t* ctx)
     return YJIT_KEEP_COMPILING;
 }
 
+// new range initialized from top 2 values
+static codegen_status_t
+gen_newrange(jitstate_t* jit, ctx_t* ctx)
+{
+    rb_num_t flag = (rb_num_t)jit_get_arg(jit, 0);
+
+    // rb_range_new() allocates and can raise
+    jit_prepare_routine_call(jit, ctx, REG0);
+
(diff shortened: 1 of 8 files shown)


yjit_codegen.c
```diff
@@ -3004,9 +3000,8 @@ gen_send_cfunc(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const
     // cfunc calls may corrupt types
     ctx_clear_local_types(ctx);
 
-    // Note: gen_send_iseq() jumps to the next instruction with ctx->sp_offset == 0
-    // after the call, while this does not. This difference prevents
-    // the two call types from sharing the same successor.
+    // Note: the return block of gen_send_iseq() has ctx->sp_offset == 1
+    // which allows for sharing the same successor.
```
jhawthorn, line 3004: 🙌

PR #197, opened 2021-09-03
Implement attr_set method calls

tenderlove:
This PR implements attr-set method calls.  We needed to fire a write barrier so it adds an "easy to call" write barrier function.  This still needs tests though (so I'll keep working on them).


yjit_codegen.c
```diff
@@ -1430,6 +1435,130 @@ gen_set_ivar(jitstate_t *jit, ctx_t *ctx, const int max_chain_depth, VALUE compt
...
+
+    // ID for the name of the ivar
+    ID id = ivar_name;
+    struct rb_iv_index_tbl_entry *ent;
+    struct st_table *iv_index_tbl = ROBJECT_IV_INDEX_TBL(comptime_receiver);
+
+    // Lookup index for the ivar the instruction loads
+    if (iv_index_tbl && rb_iv_index_tbl_lookup(iv_index_tbl, id, &ent)) {
+        uint32_t ivar_index = ent->index;
+
```
jhawthorn, line 1476: Do we also need a frozen check?

tenderlove: 🤦 ya we probably do.  I'm going to write a test for it though

tenderlove: Fixed and added tests.

yjit_codegen.c
```diff
@@ -1441,6 +1446,135 @@ gen_set_ivar(jitstate_t *jit, ctx_t *ctx, const int max_chain_depth, VALUE compt
...
+        return YJIT_END_BLOCK;
+    }
+
+    GEN_COUNTER_INC(cb, setivar_name_not_mapped);
+    return YJIT_CANT_COMPILE;
+}
+
+/*
+{
+    VALUE comptime_val_klass = CLASS_OF(comptime_receiver);
```
XrXr, line 1575: I think this comment block can go now. It's basically the code you added.

yjit_codegen.c
```diff
@@ -1441,6 +1446,135 @@ gen_set_ivar(jitstate_t *jit, ctx_t *ctx, const int max_chain_depth, VALUE compt
...
+            }
+
+            // Save recv for write barrier later
+            mov(cb, C_ARG_REGS[0], REG0);
+
+            // Get a pointer to the extended table
+            x86opnd_t tbl_opnd = mem_opnd(64, REG0, offsetof(struct RObject, as.heap.ivptr));
+            mov(cb, REG0, tbl_opnd);
+
+            // Read the ivar from the extended table
```
XrXr, line 1546: Copy pasta comment

yjit_codegen.c
```diff
@@ -1441,6 +1446,135 @@ gen_set_ivar(jitstate_t *jit, ctx_t *ctx, const int max_chain_depth, VALUE compt
...
+            mov(cb, REG1, ctx_stack_pop(ctx, 1));
+            mov(cb, ivar_opnd, REG1);
+
+            mov(cb, C_ARG_REGS[1], REG1);
+            call_ptr(cb, REG1, (void *)yjit_obj_written);
+
+            // Pop receiver if it's on the temp stack
+            // ie. this is an attribute method
+            if (!reg0_opnd.is_self) {
+              ctx_stack_pop(ctx, 1);
```
XrXr, line 1557: Wrong indent

XrXr (approved):
Since we call `rb_ivar_set` for `setinstancevariable`, I was expecting we would do the same for setter methods. We need to call a function for the write barrier anyways so that kind of makes this not a great target for optimization. It's kind of odd to spend effort and complexity points on this I guess.

tenderlove:
Ah right.  The reason I chose not to use `rb_ivar_set` is that it means we never use ivar index caches, and it seems like we'd be slower than the interpreter.  If that isn't a concern, I'm happy to change this to use `rb_ivar_set`.

maximecb:
> Ah right. The reason I chose not to use rb_ivar_set is that it means we never use ivar index caches, and it seems like we'd be slower than the interpreter. If that isn't a concern, I'm happy to change this to use rb_ivar_set.

Hmm, that is a concern. I think what Alan was saying was to model the implementation after `gen_setinstancevariable(jitstate_t* jit, ctx_t* ctx)` (or even reuse that code if possible?)

Need to remove most/all of the code from `gen_set_ivar`. No need to peek at the runtime value or do any type checking if we're calling the interpreter functions to do the ivar set.

tenderlove:
> No need to peek at the runtime value or do any type checking if we're calling the interpreter functions to do the ivar set.

We're peeking at the receiver anyway because this is a method call though, right?

> Hmm, that is a concern. I think what Alan was saying was to model the implementation after `gen_setinstancevariable(jitstate_t* jit, ctx_t* ctx)` (or even reuse that code if possible?)

As I said, I'm happy to do the refactoring if you and @xrxr think that's the best way 😄

tenderlove:
I've refactored this a bit.  We can't reuse the existing instance variable setting code because the inline cache for methods has a different shape than the inline cache for regular ivar sets.  I've pulled out a function that forces getting an index for an ivar, then caches the ivar index in the machine code.

I think we might be able to do the same technique with regular ivars but we'd have to peek at the runtime type and I don't think we're doing that right now (AFAICT)

vm_insnhelper.c
```diff
@@ -1369,6 +1369,24 @@ rb_vm_setinstancevariable(const rb_iseq_t *iseq, VALUE obj, ID id, VALUE val, IV
...
+
+    if (UNLIKELY(index >= ROBJECT_NUMIV(obj))) {
+        rb_init_iv_list(obj);
+    }
+    VALUE *ptr = ROBJECT_IVPTR(obj);
+    RB_OBJ_WRITE(obj, &ptr[index], val);
+    RB_DEBUG_COUNTER_INC(ivar_set_ic_hit);
+    return val; /* inline cache hit */
+}
+
```
maximecb, line 1389: Should add a comment explaining what this function does. Otherwise looks pretty good :)

tenderlove: Added :)

PR #198, opened 2021-09-03
Add codegen for rb_true and rb_false

jhawthorn:
These are used by `.nil?` and therefore `opt_nil_p`.

For `nil.nil?` we generate:

```
== disasm: #<ISeq:block in <main>@(eval):1 (1,5)-(1,17)> (catch: FALSE)
0000 putnil                                                           (   1)[LiBc]
0001 opt_nil_p                              <calldata!mid:nil?, argc:0, ARGS_SIMPLE>[CcCr]
0003 leave                                  [Br]

== BLOCK 1/3: 7 BYTES, ISEQ RANGE [0,3) ========================================
  ; putnil
  55ba551d105d:  mov    qword ptr [rbx], 8
== BLOCK 2/3: 10 BYTES, ISEQ RANGE [1,3) =======================================
  ; opt_nil_p
  55ba551d1064:  mov    rax, qword ptr [rbx]
  ; nil? == true
  55ba551d1067:  mov    qword ptr [rbx], 0x14
```

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -2783,6 +2783,28 @@ jit_rb_obj_not(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const 
     return true;
 }
 
+// Codegen for rb_true()
+static bool
+jit_rb_true(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const rb_callable_method_entry_t *cme, rb_iseq_t *block, const int32_t argc)
+{
+    ADD_COMMENT(cb, "nil? == true");
+    ctx_stack_pop(ctx, 1);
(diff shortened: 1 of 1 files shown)


yjit_codegen.c
```diff
@@ -4102,4 +4124,7 @@ yjit_init_codegen(void)
     yjit_method_codegen_table = st_init_numtable();
 
     yjit_reg_method(rb_cBasicObject, "!", jit_rb_obj_not);
+
+    yjit_reg_method(rb_cNilClass, "nil?", jit_rb_true);
+    yjit_reg_method(rb_mKernel, "nil?", jit_rb_false);
```
maximecb, line 4129: What's the pattern for Kernel and nil? in terms of Ruby code?

PR #199, opened 2021-09-03
Stamp out some out of memory crashes 

XrXr:
This change makes it so that we don't crash with rb_bug() when we run
out of executable memory while generating code for a new entry point
or when hitting a lazy stub.

A new API, `YJIT.simulate_out_of_mem!`, is used to test for out of
memory conditions.

If we run out while generating a new entry point, we can handle it by
not installing an entry point. Handling running out while compiling a
lazy stub is more involved.

An exit for the stub target is added to the the piece of code that calls
branch_stub_hit() and used when branch_stub_hit() returns NULL. Unlike
before, being able to generate new code is no longer required for
lazy stubs to function correctly.

A situation not handled by this change is when we need to invalidate a
block while we are out of memory. We generate new stubs when we
invalidate and that is not handled for now.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1928,3 +1928,44 @@ assert_equal '42', %q{
 
   ractor.take
 }
+
+# Test generating new code while out of memory
+assert_equal 'ok', %q{
+  def foo
+    :ok
+  end
+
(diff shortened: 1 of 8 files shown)


maximecb:
I appreciate the intention, but might not be the best time to merge this. I opened a PR for code pages which might touch a lot of the same code.

PR #201, opened 2021-09-06
Allow calling variadic cfuncs with many args

jhawthorn:
We have a check to ensure we don't have to push args on the stack to call a cfunc with many args (we probably could support this, but will save that for a future PR). However we never need to use the stack for variadic cfuncs, so we shouldn't care about the number of arguments.

Found this side exit from one of the parsers in main in https://github.com/Shopify/yjit-bench/pull/40


maximecb:
So it uses the Ruby stack for variadic C functions?

jhawthorn:
Yes, we pass a pointer to the arguments (which happen to be from the Ruby stack)

https://github.com/Shopify/yjit/blob/main/yjit_codegen.c#L3005-L3011

PR #203, opened 2021-09-06
Implement getspecial

jhawthorn:
This implements `getspecial` as used for the special regex "globals".

I implemented this using `rb_backref_get` for simplicity, which is probably a bit slower than finding the SVAR through the CFP we already have.


yjit_codegen.c
```diff
@@ -3770,6 +3770,70 @@ gen_toregexp(jitstate_t* jit, ctx_t* ctx)
...
+static codegen_status_t
+gen_getspecial(jitstate_t *jit, ctx_t *ctx)
+{
+    rb_num_t key = jit_get_arg(jit, 0);
+    rb_num_t type = jit_get_arg(jit, 1);
+
+    if (type == 0) {
+        // not yet implemented
+        return YJIT_CANT_COMPILE;
+    } else if (type & 0x01) {
```
maximecb, line 3782: Could use a comment explaining where the 0x01 comes from

maximecb:
👀 just minor changes in terms of commenting and we can merge this.

jhawthorn:
Added some comments. There's unfortunately not too much clarification to add. The purpose of these values is the thing they do 😅.

PR #204, opened 2021-09-07
Exit if YJIT and MJIT are both enabled

tenderlove:
YJIT and MJIT can't be running in the same process otherwise they'll
clobber each other.  We should show an error and exit if they're both
enabled.


ruby.c
```diff
@@ -1824,13 +1824,18 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
...
-    if (opt->features.set & FEATURE_BIT(yjit))
-        rb_yjit_init(&opt->yjit);
 #if USE_MJIT
     if (opt->features.set & FEATURE_BIT(jit)) {
         opt->mjit.on = TRUE; /* set mjit.on for ruby_show_version() API and check to call mjit_init() */
     }
 #endif
+    if (opt->features.set & FEATURE_BIT(yjit)) {
+        if (opt->mjit.on) {
+            rb_warn("MJIT and YJIT are both enabled at the same time. Exiting");
```
maximecb, line 1834:
```suggestion
            rb_warn("MJIT and YJIT cannot both be enabled at the same time. Exiting");
```

maximecb:
@casperisfine once this is merged can you bump the deployed YJIT version to include this?

tenderlove:
Looks like we're failing [this test](https://github.com/ruby/spec/blob/b1e93a2b4238c6510641a0a1f112265e5c58f5f7/command_line/feature_spec.rb#L46-L53) now.  I'm not 100% sure what to do about it since it's in `ruby/spec` and I don't think it's a good idea if we fork and modify another repo.

The issue is that you can do `ruby --enable-all` and that ends up enabling both MJIT and YJIT.  Does anyone have any suggestions for how to proceed?

paracycle:
> The issue is that you can do `ruby --enable-all` and that ends up enabling both MJIT and YJIT. 

Can we turn off MJIT in that case, if YJIT is in play? Would that solve the problem?

maximecb:
Yeah I would make it so `ruby --enable-all` enables YJIT and disables MJIT.

Random aside: what kind of stupid option is `--enable-all`....? Why would anybody ever want that???

test/ruby/test_rubyoptions.rb
```diff
@@ -164,9 +164,9 @@ def test_copyright
 
   def test_enable
     if JITSupport.supported?
-      assert_in_out_err(%w(--enable all -e) + [""], "", [], [])
-      assert_in_out_err(%w(--enable-all -e) + [""], "", [], [])
-      assert_in_out_err(%w(--enable=all -e) + [""], "", [], [])
+      assert_in_out_err(%w(--enable all --disable-yjit -e) + [""], "", [], [])
+      assert_in_out_err(%w(--enable-all --disable-yjit -e) + [""], "", [], [])
+      assert_in_out_err(%w(--enable=all --disable-yjit -e) + [""], "", [], [])
```
maximecb, line 169: This probably needs to be reverted since we want `--enable-all` to work as-is.

test/ruby/test_rubyoptions.rb
```diff
@@ -229,16 +229,16 @@ def test_version
 
     if JITSupport.supported?
       [
-        %w(--version --jit),
-        %w(--version --enable=jit),
-        %w(--version --enable-jit),
+        %w(--version --disable-yjit --jit),
+        %w(--version --disable-yjit --enable=jit),
+        %w(--version --disable-yjit --enable-jit),
```
maximecb, line 234: There's a couple of places where you added `--disable-yjit` in this file. Just double-checking, is that necessary?

tenderlove: Ya, I think it's necessary.  I think those tests are specifically testing that we can enable MJIT.

PR #212, opened 2021-09-10
Fix avg_len_in_yjit

XrXr:
We weren't counting completing an entire method in YJIT as exits so the
avg_len_in_yjit for

    ./miniruby --yjit-call-threshold=1 --yjit-stats -e'def foo; end; foo'

was infinite.


noahgibbs:
Ah, nice. I'll update this in the yjit-metrics YJIT exit stats as well.

PR #215, opened 2021-09-12
Merge upstream

jhawthorn:
Testing this with a fix (that I'll submit separately if passing).

I think the bug is from my opt_eq PR.


maximecb:
Thanks a lot for taking a look at this John 👌

PR #218, opened 2021-09-14
Include errno message in mmap failure output

casperisfine:
It might help figure out why it is failing.


noahgibbs:
Cool. Yeah, the man page only notes that perror has a bit of a race condition with stderr, which I think just means output can arrive in different orders (which is fine.) So yeah, perror addresses my concern.

noahgibbs (approved):

PR #219, opened 2021-09-14
Remove a couple of unnecessary memory stores

XrXr:
See commits.


yjit_codegen.c
```diff
@@ -352,9 +352,6 @@ yjit_gen_exit(VALUE *exit_pc, ctx_t *ctx, codeblock_t *cb)
         mov(cb, member_opnd(REG_CFP, rb_control_frame_t, sp), REG_SP);
     }
 
-    // Update the CFP on the EC
-    mov(cb, member_opnd(REG_EC, rb_execution_context_t, cfp), REG_CFP);
-
```
maximecb, line 357: Curious why this doesn't need too be set, does the interpreter set it after we exit?

XrXr: We set it when we do a method call. It's why in `gen_leave()` it's just `ec->cfp--;`. When we enter generated code it's already set for us before entering.

PR #221, opened 2021-09-14
Fix warnings about redefining YJIT_STATS

XrXr:
Follow up for ecb5b383a0c17550b9b27663005049ddac871edb. Now that
YJIT_STATS is defined in yjit.h, it shoudl be the only place that
defines it.


maximecb:
Thank you Alan 🙏

PR #222, opened 2021-09-14
Specialize opt_aset based on received types

jhawthorn:
This replaces the existing implementation of `opt_aset`, which would call the implementation from the interpreter, which only worked for Array and Hash. This new implementation instead specializes and uses chain guards to call either `rb_ary_store` or `rb_hash_aset` or fall back to a standard method call.

Previously: https://github.com/Shopify/yjit/pull/220#issuecomment-919379462
I suspect other benchmarks will also feature this (Rails uses Concurrent::Map pretty heavily)

### DIFF
--- a/yjit_codegen.c
+++ b/yjit_codegen.c
@@ -2187,32 +2187,81 @@ VALUE rb_vm_opt_aset(VALUE recv, VALUE obj, VALUE set);
 static codegen_status_t
 gen_opt_aset(jitstate_t *jit, ctx_t *ctx)
 {
-    // Save the PC and SP because the callee may allocate
-    // Note that this modifies REG_SP, which is why we do it first
-    jit_prepare_routine_call(jit, ctx, REG0);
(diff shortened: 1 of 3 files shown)


yjit_codegen.c
```diff
@@ -2182,37 +2182,84 @@ gen_opt_aref(jitstate_t *jit, ctx_t *ctx)
...
+        sar(cb, C_ARG_REGS[1], imm_opnd(1)); // FIX2LONG(key)
+        mov(cb, C_ARG_REGS[2], val);
+
+        // We might allocate or raise
+        jit_prepare_routine_call(jit, ctx, REG0);
+
+        call_ptr(cb, REG0, (void *)rb_ary_store);
+
+        // rb_ary_store returns void
+	// stored value should still be on stack
```
maximecb, line 2226:
```suggestion
	 // stored value should still be on stack
```

maximecb:
Thank you John, this looks good! :)

XrXr (approved):

PR #224, opened 2021-09-15
Add counters for tracking invalidations

XrXr:


maximecb (commented):
Thanks a lot Alan! :)

PR #228, opened 2021-09-17
Fix opt_aset comptime_key check

jhawthorn:
This had a typo where it was checking `val` instead of `key` 🤦‍♂️

Adds a test which previously had an assertion fail in `jit_guard_known_klass`


maximecb (approved):
Good catch 😆 

What would we do without you.

PR #230, opened 2021-09-17
Implement invokebuiltin

jhawthorn:
We've previously implemented `opt_invokebuiltin_delegate` (and `&_leave` ) but not the non-optimized `invokebuiltin`, which takes arguments on the stack instead of directly from caller locals.

This instruction is used (among other things) for assignment to `Struct` members, which is used somewhat heavily in the mail benchmark (13% of remaining exits, but in important places so this should significantly increase our coverage).

This also fixes a check between argc and `NUM_C_ARG_REGS` in `opt_invokebuiltin_delegate` which was too strict (these are both counting the number of arguments).


yjit_codegen.c
```diff
@@ -4144,7 +4177,7 @@ gen_opt_invokebuiltin_delegate(jitstate_t *jit, ctx_t *ctx)
     const struct rb_builtin_function *bf = (struct rb_builtin_function *)jit_get_arg(jit, 0);
     int32_t start_index = (int32_t)jit_get_arg(jit, 1);
 
-    if (bf->argc + 2 >= NUM_C_ARG_REGS) {
+    if (bf->argc + 2 > NUM_C_ARG_REGS) {
         return YJIT_CANT_COMPILE;
     }
```
maximecb, line 4182: Same here :)

PR #232, opened 2021-09-20
Add jit_obj_info_dump debugging method

jhawthorn:
This adds `jit_obj_info_dump` which generates runtime code to call `rb_obj_info_dump` on the provided operand. This should provide friendlier debugging output than `print_ptr` when inspecting a Ruby VALUE.

`rb_obj_info_dump` shouldn't allocate (it can be used in GC debugging) so this should be safe to call wherever.


maximecb:
Thanks John!

PR #236, opened 2021-09-21
Allow getinlinecache to be recompiled from entrypoint

jhawthorn:
Based on top of https://github.com/Shopify/yjit/pull/233

Previously when the first instruction of a method was getinlinecache, and the method was compiled as a jit_func entrypoint (vs. being compiled from `gen_send_iseq`), if the cache was not up to date we would never get the chance to recompile it when the cache was updated.

This PR make two changes to improve this:

First is to reset `iseq->body->total_calls` to `0` when setting `jit_func` to `NULL`. This allows us to attempt recompilation through the `mjit_exec` path. Previously this would never have been reattempted. This may possibly help other instructions invalidated at the start of methods.

Second is to explicitly generate a side exit when the inline cache is outdated instead of relying on `YJIT_CANT_COMPILE`. This ensures that `gen_entry_point` doesn't return `NULL`.

This may feel like a bit of a hack, but I think this is a specific quirk of `getinlinecache`. Most instructions we don't expect to go from uncompilable to compiled. It could also make sense for this instruction to push nil and side exit to the **next** instruction as a sort of "successful side exit".


XrXr:
I feel like this is not big enough of a problem to add even more special handling for `opt_getinlinecahce`. For threshold > 1, we are always going to have a warm cache when we compile if the instruction sits at entry point.

Also there is a problem with this in that if the cache has a cref, we would leave a block in there that never gets invalidated, which leaves us worse off than before, where we would not install the entry point.

Resetting `total_calls` feels like a hack and may have consequences beyond handling this. It's currently naturally selecting for entry points that haven't been invalidated.

maximecb:
I feel like this is a whole separate idea, but I kind of like the idea of resetting the call counter to zero if we invalidate a block to give the method another chance at compilation. It's nice because it means we don't even spend time compiling a stub, the method has to "prove itself useful" again to be compiled. Just wondering if there's a risk that this could produce an infinite loop of compilation and invalidation in some cases.

XrXr:
> do you think there's merit to resetting the call counter so we get a shot at recompiling methods?

I don't really have a strong opinion about it. If it helps some specific benchmark, sure why not. I feel like maybe we don't need to worry about the entry point that much, since we can do calls. Invalidation should be rare anyways, and sometimes the start of the method isn't something that could be invalidated, like `putself`.

XrXr:
Because of `defer_compilation`, I think the send and `putself` ends up in different blocks.

jhawthorn:
> For threshold > 1, we are always going to have a warm cache when we compile if the instruction sits at entry point.

We might always have a cache with a value, but it won't necessarily be valid.

> Also there is a problem with this in that if the cache has a cref, we would leave a block in there that never gets invalidated, which leaves us worse off than before, where we would not install the entry point.

This doesn't change handling of cref.

> Resetting `total_calls` feels like a hack and may have consequences beyond handling this. It's currently naturally selecting for entry points that haven't been invalidated.

I think "invalidating" is an indication that how we need to handle the instruction may have changed, not that how we handle the instruction is now invalid (which we'll detect when we try to recompile). If there's an issue invalidating the first instruction we're likely going to hit it on non-first instructions. I think it's surprising behaviour that the first instruction in an iseq is handled so differently than the others.

PR #240, opened 2021-09-21
Add a slowpath for opt_getinlinecache

XrXr:
Before this change, when we encounter a constant cache that is specific
to a lexical scope, we unconditionally exit. This change falls back to
the interpreter's cache in this situation.

This should help constant expressions in `class << self`, which is popular
at Shopify due to the style guide.

This change relies on the cache being warm while compiling to detect the
need for checking the lexical scope for simplicity.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -289,6 +289,34 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_opt_getinlinecache_slowpath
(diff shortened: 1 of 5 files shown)


XrXr:
Another thing we could do is to leave the generic version when the cache is cold. This way we would be able to handle `cref` sites even when we have cold cache when we compile. The tradeoff is that we would general more code overall for simple sites that compile down to simple push.

maximecb:
> Another thing we could do is to leave the generic version when the cache is cold. This way we would be able to handle cref sites even when we have cold cache when we compile.

I guess we can try deploying this and see how many side-exits are left. If it's common, might be worth generating somewhat bigger code to get more coverage.

PR #241, opened 2021-09-22
Add counters for version invalidation reasons

XrXr:
I noticed that there were two st_table iterators that do exactly the
same thing so I merged them into one. The new iterators does the
counting and expects the caller to pass in the reason.

### DIFF
--- a/yjit_iface.c
+++ b/yjit_iface.c
@@ -318,11 +318,13 @@ static const rb_data_type_t yjit_root_type = {
     0, 0, RUBY_TYPED_FREE_IMMEDIATELY
 };
 
+// st_table iterator for invalidating blocks that are keys to the table.
 static int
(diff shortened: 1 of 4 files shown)


yjit_iface.c
```diff
@@ -375,7 +411,7 @@ rb_yjit_cme_invalidate(VALUE cme)
         st_table *block_set = (st_table *)blocks;
 
         // Invalidate each block
-        st_foreach(block_set, block_set_invalidate_i, 0);
+        st_foreach(block_set, invalidate_block_key_i, (st_data_t)INVAL_METHOD_LOOKUP);
```
maximecb, line 414: The answer is probably no, but couldn't you just ask for the length of the block set and increment the counter directly?

XrXr: Ah yeah that's better.

maximecb (approved):
Looks great 👌 Thank you Alan :)

PR #253, opened 2021-09-28
YJIT should be inactive by default

noahgibbs:
Remove YJIT feature bit from default set.
Stop deactivating MJIT when enabling all.

I don't think we need to change the two command line parameters for YJIT (--yjit, --enable-yjit) since both have parallels for MJIT (--jit, --enable-jit).


maximecb:
Might need to audit the workflows in .github to make sure we're testing YJIT enough if it's disabled by default. We do want to make sure to test it on multiple platforms and to run tests like test-bundler-parallel. I'll ask @XrXr to review.

maximecb (commented):
This disables YJIT when --enable-all is set, but YJIT is still enabled by default I think.

noahgibbs:
Hm. This acts for me like YJIT is disabled by default (e.g. no YJIT mentioned in ruby -v, YJIT.runtime_stats is nil.) But I may be missing something.

noahgibbs:
It's working for me, but Maxime was worried that it didn't work - I wasn't 100% sure I was testing the right thing. But it seems to work in the ways I've tested.

maximecb:
It's the github workflows I'm concerned about. If YJIT is disabled by default, that means we're running a lot less tests with YJIT enabled, which is not good. We're a lot more likely to let bugs creep in.

Let's make sure that we're testing `test-bundled-gems` and `test-bundler-parallel` with YJIT enabled and `--yjit-call-threshold=1`.

I think it might also be wise to make sure that the YJIT options workflow runs on both Ubuntu and macOS.

noahgibbs:
Okay. I've added the additional testing. Waiting on Maxime to make sure this adequately addresses her concerns.

maximecb:
Looks like the GitHub workflows are refusing to run, maybe a parse error?

maximecb:
Maybe we should take this opportunity to rename the "YJIT options" workflow just "YJIT" for upstreaming.

noahgibbs:
Okay. If we don't want bug reports on Windows, should we add a warning for YJIT on Windows before upstreaming?

I would think we were already testing YJIT on Windows on GitHub CI. The windows.yml workflow doesn't look disabled, or like it disables YJIT, at a glance. We turned off mingw, but (as far as I can tell) not compiling with Visual Studio.

XrXr:
I don't know how upstream will like it. I think we can send it and see what people say.

PR #255, opened 2021-09-28
send iseq kwargs

kddnewton:
This accounts for the case where you're sending keyword arguments to a method that specifies keyword arguments and you're passing a value for every one of them. It does not yet handle optionals.


yjit_codegen.c
```diff
@@ -3384,9 +3385,114 @@ gen_send_iseq(jitstate_t *jit, ctx_t *ctx, const struct rb_callinfo *ci, const r
...
+                    if (callee_kwarg == caller_kwargs[swap_idx]) {
+                        x86opnd_t swap = ctx_stack_opnd(ctx, argc - 1 - swap_idx);
+                        x86opnd_t kwarg = ctx_stack_opnd(ctx, argc - 1 - kwarg_idx);
+
+                        // First we're going to generate the code that is going
+                        // to perform the actual swapping at runtime.
+                        mov(cb, REG1, swap);
+                        mov(cb, R9, kwarg);
+                        mov(cb, swap, R9);
+                        mov(cb, kwarg, REG1);
```
maximecb, line 3468: I wonder if we could make this a little more compact by using an `xchg` instruction?
https://www.felixcloutier.com/x86/xchg

Just a remark, I understand if that waits until another PR. Correctness first.

maximecb (commented):
Good job Kevin! The code makes sense. We might want to factor some of it into another method because `gen_send_iseq` is going to be huuuuge, but that can wait until after missing features are added and tests are passing.

yjit_codegen.c
jhawthorn: https://www.agner.org/optimize/ recommends against xchg (though the trade off might be worth it for instruction size)

kddnewton: I tried using the `xchg` but we don't have it working with a memory operand and a register operand at the moment. Let's pair on that tomorrow I'd love to use that instruction.

XrXr: Ah good catch. We probably shouldn't use it because with a memory operand it always has the LOCK prefix. From Intel's manual:

> If a memory operand is referenced, the processor’s locking protocol is automatically implemented for the duration of the exchange operation, regardless of the presence or absence of the LOCK prefix or of the value of the IOPL.

maximecb: Aaah if it does locking then probably best not to use it.

PR #262, opened 2021-10-01
Mockup PR for Big Rebase

noahgibbs:
This should let us see what the GitHub UI will show for the real PR later.

This PR is relative to the rebase_upstream_ruby branch, which exists only to show the diff properly. When we get rid of this mockup PR, we can also get rid of that branch.

Most recent upstream commit: 57bf354c9a878bb67c294408400fd029f9b5a353

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -2160,3 +2160,18 @@ assert_equal '[2]', %q{
 
   5.times.map { default_expression(value: 2) }.uniq
 }
+
+# attr_reader on frozen object
+assert_equal 'false', %q{
+  class Foo
+    attr_reader :exception
+
+    def failed?
+      !exception.nil?
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


iseq.c
```diff
@@ -3213,14 +3220,33 @@ rb_vm_insn_addr2insn(const void *addr)
...
+        int opcode = e->insn;
+        if (addr == e->trace_encoded_insn) {
+            opcode += VM_INSTRUCTION_SIZE/2;
+        }
+        return opcode;
+    }
+
+    rb_bug("rb_vm_insn_addr2opcode: invalid insn address: %p", addr);
+}
+
```
maximecb, line 3249: @XrXr I think we might be able to revert much of these changes to iseq.c since we changed to use the MJIT hook for integration.

maximecb: Could you comment on other bits we might not need in iseq.c?

PR #263, opened 2021-10-01
YJIT docs update for upstreaming

noahgibbs:
Add a YJIT hacking doc with some internals
Change the README text a bit to reflect its current condition


doc/yjit/yjit_hacking.md
```diff
@@ -0,0 +1,79 @@
...
+
+YJIT’s basic purpose is to take ISEQs and generate machine code.
+
+Documentation on each Ruby bytecode can be found in insns.def.
+
+YJIT uses those bytecodes as the “Basic Blocks” in Lazy Basic Block Versioning (LBBV.) For more deep details of LBBV:
+
+* Basic Block Versioning (whitepaper) - https://arxiv.org/abs/1411.0352
+* An Extension of BBV using Object Shapes (whitepaper) - https://arxiv.org/abs/1507.02437
+* Basic Block Versioning talk at ECOOP 2015 - https://www.youtube.com/watch?v=S-aHBuoiYE0
```
maximecb, line 13: We have links to both of these papers and accompanying talks in the current README.md

doc/yjit/yjit_hacking.md
```diff
@@ -0,0 +1,79 @@
...
+
+When YJIT is generating code, it needs a code pointer. In many cases it needs two, usually called “cb” (codeblock) and “ocb” (out-of-line codeblock.)
+
+cb is for “inlined” normal code and ocb is for “outline” code such as exits. Inlined code is normal generated code for Ruby operations, while outlined code is for unusual and error conditions, such as encountering an unexpected parameter type and exiting to the interpreter.
+
+Simple inline code for a method body runs faster, so it's useful to remove unexpected exits from the control flow. An exception or unsupported operation will cause YJIT to generate out-of-line code to handle it.
+
+If you search for ocb in yjit_codegen.c, you can see some places where out-of-line code is generated.
+
+YJIT statistics are only gathered when RUBY_DEBUG or YJIT_STATS is true. In some cases the code to increment YJIT statistics will be generated out-of-line, especially if those statistics are gathered when a side exit happens.
```
maximecb, line 60: The purpose of the outlined code block is to keep things we believe are going to be infrequent somewhere else. That way we can keep the code in the inline block more linear and compact. Linear code, with as few branches as possible, is more easily predicted by the CPU.

maximecb:
We should probably move our whole current `README.md` into `doc/yjit/README.md` to preserve the content but avoid conflicts with the upstream README.md.

noahgibbs:
In the rebase, I've already moved README.md to doc/yjit/yjit.md, which was what you suggested elsewhere. I'd be fine with calling it doc/yjit/README.md too.

For the README in the rebase, there are some interesting questions about what we want to do with file history (unless we just do a big squash of all the commits.)

maximecb:
Thanks for putting this together Noah 🙏

PR #270, opened 2021-10-04
Put YJIT into a single compilation unit

XrXr:
For upstreaming, we want functions we export either prefixed with "rb_"
or made static. Historically we haven't been following this rule, so we
were "leaking" a lot of symbols as `make leak-globals` would tell us.

This change unifies everything YJIT into a single compilation unit,
yjit.o, and makes everything unprefixed static to pass `make leak-globals`.
This manual "unified build" setup is similar to that of vm.o.

A lot of the unprefixed non-static functions are coming from the assembler,
where each opcode gets their own function. To make these functions static
while allowing uses outside of Ruby that want these functions to have
external linkage, I added a new macro called `YJIT_ASM_BUNDLED` to control
the linkage. `yjit_asm_tests.c` is an example of this use case.

Other changes of note:
  - switched to MJIT_SYMBOL_EXPORT_BEGIN which indicates stuff as being
    off limits for native extensions
  - the first include of each YJIT file is change to be "internal.h"
  - undefined MAP_STACK before explicitly redefining it since it
    collide's with a definition in system headers. Consider renaming?


maximecb:
Can we just use the same mechanism in the asm tests (compile that in to a single unit) and not need `YJIT_ASM_BUNDLED` ?

XrXr:
> switched to MJIT_SYMBOL_EXPORT_BEGIN which indicates stuff as being
off limits for native extensions

Since we don't support calling YJIT code from MJIT, it might be possible to make it so that we don't export these at all. Off topic here though.

XrXr:
> Can we just use the same mechanism in the asm tests (compile that in to a single unit) and not need `YJIT_ASM_BUNDLED`?

Sounds good. Does that mean it's time to stop trying to use CRuby stuff in the assembler?

maximecb:
> Sounds good. Does that mean it's time to stop trying to use CRuby stuff in the assembler?

I suppose we can.

maximecb:
Yes we can do that if we want to 👍

maximecb (approved):

XrXr:
The CI failure is from #271

PR #276, opened 2021-10-07
Remove yjit_backend to help with upstreaming

XrXr:
The new backend isn't used at the moment and adds to our diff against
upstream so remove it for now. We can reverse the removal later with git
history.


noahgibbs:
The backend commits shouldn't be in the rebased branch - I left them out on purpose.

XrXr:
Right, but the build is breaking on the rebase branch because the removal is incomplete.

maximecb:
Sure thing. @XrXr can you make sure to leave these on a branch or have a tag for them?

XrXr:
Okay, I made a snapshot of main at https://github.com/Shopify/yjit/tree/new-backend.

PR #277, opened 2021-10-07
Show +YJIT in version string and RUBY_DESCRIPTION

XrXr:
While working on this, I found some tests that fail when `RUBY_YJIT_ENABLE` is set.
Maybe setting that variable should be part of the build matrix too? 

---

There might be code out there that expect `ruby -v` to print only one
line. Since MJIT shows +JIT in `ruby -v` and RUBY_DESCRIPTION, let's
show +YJIT.

The crash report doesn't show anything about MJIT, so adjust the test.
The "test_ruby_version" test was unaware of RUBY_YJIT_ENABLE and so
was failing when the variable is set and inherited into the children
processes it spawns. Explicitly unset the variable in the test.

### DIFF
--- a/.github/workflows/yjit-ubuntu.yml
+++ b/.github/workflows/yjit-ubuntu.yml
@@ -4,7 +4,7 @@ jobs:
   make:
     strategy:
       matrix:
-        test_task: ["check", "test-bundled-gems"] # "test-bundler-parallel",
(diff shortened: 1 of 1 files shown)


XrXr:
Working on this turned out to reveal a few issues with our new CI setup now that YJIT is off by default. Does everything look good? cc @noahgibbs

PR #282, opened 2021-10-15
Bugfix: Don't modify object in yjit_force_iv_index

jhawthorn:
Previously this method would set the instance variable to `Qundef` as a convenient way to ensure that the variable was defined on the class. However if the object was frozen this would incorrectly raise a `FrozenError` trying to modify the object.

This commit changes yjit_force_iv_index to not modify the object and only modify the class iv table.


jhawthorn:
Updated with feedback

